PR #32, opened 2018-05-24
Multi UWerr

urbach:
this will allow to perform the UW analysis on multiple quantities in a single call, i.e. UWerr for vector valued functions.
For certain quantities this reduces runtime significantly.

However, it will need a thorough check, I might have missed points where `uwerr` is called.


urbach:
I tried to direct for this, but I'll check again. Our maybe you have a quick usecase?

On 25 May 2018 10:24:58 CEST, Bartosz Kostrzewa <notifications@github.com> wrote:
>kostrzewa commented on this pull request.
>
>
>
>>    }
>   \item{data}{
>-    array of data to be analysed. It must be of dimension (Nalpha x N)
>-    (i.e. Nalpha rows and N columns),
>+    array of data to be analysed. It must be of dimension (N x Nalpha)
>
>So this breaks ```uwerr.cf``` right now, correct?

Carsten Urbach, www.carsten-urbach.eu

R/UWerr.R
```diff
@@ -344,45 +354,68 @@ uwerrderived <- function(f, data, nrep, S=1.5, pl=FALSE, ...) {
...
+      Wopt <- x$Wopt[[index]]
+      Wmax <- x$Wmax[[index]]
+    }
     if(x11) if(interactive() && (grepl(pattern="X11", x=names(dev.cur()), ignore.case=TRUE) || grepl(pattern="null", x=names(dev.cur()), ignore.case=TRUE))) {
               X11()
             }
-    plotwitherror(c(0:x$Wmax),GammaFbb[1:(x$Wmax+1)],
-                  Gamma.err[1:(x$Wmax+1)], ylab="Gamma(t)", xlab="t", main=main)
-    abline(v=x$Wopt+1)
+    plotwitherror(c(0:Wmax),GammaFbb[1:(Wmax+1)],
```
kostrzewa: could we please use named arguments whenever possible? I think we need to make additional modifications to ```plotwitherror``` in the coming months and it would be good if stuff didn't break as a result...

R/UWerr.R
```diff
@@ -240,98 +239,109 @@ uwerrderived <- function(f, data, nrep, S=1.5, pl=FALSE, ...) {
...
-    ##stop("Gamma pathological, estimated error^2 <0\n")
-    warning("Gamma pathological, estimated error^2 <0\n")
-  }
-  GammaFbb <- GammaFbb+CFbbopt/N # bias in Gamma corrected
-  CFbbopt <- GammaFbb[1] + 2*sum(GammaFbb[2:(Wopt+1)]) #refined estimate
-  sigmaF <- sqrt(abs(CFbbopt)/N) #error of F
-  rho <- GammaFbb/GammaFbb[1]
-  tauintFbb <- cumsum(rho)-0.5 #normalised autocorrelation
+    
+    res$value[i] <- Fbb[i]
```
kostrzewa: This breaks ```analysis_online```

kostrzewa:
This messes with quite a lot of stuff. I think we can use the opportunity to design some integration tests.

urbach:
> This messes with quite a lot of stuff. I think we can use the opportunity to design some integration tests.

But most of it is in scripts outside of hadron/R then?

> 
> -- 
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/HISKP-LQCD/hadron/pull/32#issuecomment-391983225
--
Carsten Urbach
e-mail: curbach@gmx.de
        urbach@hiskp.uni-bonn.de
Fon   : +49 (0)228 73 2379
skype : carsten.urbach

urbach:
> > But most of it is in scripts outside of hadron/R then?
> 
> No, ```analysis_online``` and ```analysis_gradient_flow``` as well as ```uwerr.cf``` have been part of hadron for a while now.

yes, I know. But everything in hadron/R I tried to modify accordingly
(of course, I oversaw things as usual). I'll have a look again.

--
Carsten Urbach
e-mail: curbach@gmx.de
        urbach@hiskp.uni-bonn.de
Fon   : +49 (0)228 73 2379
skype : carsten.urbach

urbach:
btw: `uwerrprimary` is not affected...

> > > But most of it is in scripts outside of hadron/R then?
> > 
> > No, ```analysis_online``` and ```analysis_gradient_flow``` as well as ```uwerr.cf``` have been part of hadron for a while now.
> 
> yes, I know. But everything in hadron/R I tried to modify accordingly
> (of course, I oversaw things as usual). I'll have a look again.
> 
> --
> Carsten Urbach
> e-mail: curbach@gmx.de
>         urbach@hiskp.uni-bonn.de
> Fon   : +49 (0)228 73 2379
> skype : carsten.urbach
> URL: http://www.carsten-urbach.eu


--
Carsten Urbach
e-mail: curbach@gmx.de
        urbach@hiskp.uni-bonn.de
Fon   : +49 (0)228 73 2379
skype : carsten.urbach

kostrzewa:
> btw: `uwerrprimary` is not affected...

Okay, so ```uwerr.cf``` is fine then.

urbach:
> 
> 
> 
> > @@ -65,6 +68,15 @@ uwerrderived(f, data, nrep, S=1.5, pl=FALSE, ...)
>    \item{Wopt}{
>      value of optimal cut-off for the Gamma function integration
>    }
> +
> +  In case of a derived observable (\code{uwerrderived}), i.e. if a
> +  function is specified, the above objects are contained in a list
> +  called \code{res}.
> +
> +  In both cases the return object containes
> +  \item{xxx}{
> 
> Was this intentional?

well... (hust...)
--
Carsten Urbach
e-mail: curbach@gmx.de
        urbach@hiskp.uni-bonn.de
Fon   : +49 (0)228 73 2379
skype : carsten.urbach

urbach:
> 
> 
> 
> > -    warning("Windowing condition failed!")
> -    Wopt <- Wmax
> -  }
> -  
> -  CFbbopt <- GammaFbb[1] + 2*sum(GammaFbb[2:(Wopt+1)])
> -  if(CFbbopt <= 0) {
> -    ##stop("Gamma pathological, estimated error^2 <0\n")
> -    warning("Gamma pathological, estimated error^2 <0\n")
> -  }
> -  GammaFbb <- GammaFbb+CFbbopt/N # bias in Gamma corrected
> -  CFbbopt <- GammaFbb[1] + 2*sum(GammaFbb[2:(Wopt+1)]) #refined estimate
> -  sigmaF <- sqrt(abs(CFbbopt)/N) #error of F
> -  rho <- GammaFbb/GammaFbb[1]
> -  tauintFbb <- cumsum(rho)-0.5 #normalised autocorrelation
> +    
> +    res$value[i] <- Fbb[i]
> 
> This breaks ```analysis_online```

I don't think so, because if `length(res$value) = 1`, then `res$value`
and `res$value[1]` can both be used and give the same result. And
`length(res$value) = 1` is guaranteed if the function `f` is a scalar
function.

Of course, in `analysis_online` it looks then not so clenaly
programmed, but it should not break, should it?

--
Carsten Urbach
e-mail: curbach@gmx.de
        urbach@hiskp.uni-bonn.de
Fon   : +49 (0)228 73 2379
skype : carsten.urbach

kostrzewa:
This seems to break ```onlinemeas```, in particular ```effectivemass``` called there and perhaps others.

kostrzewa:
```
Error in uwerrderived(f = effmass2, data = Z[t:(t + 1), ], S = S, pl = F,  : 
  Error, inconsistent N and nrep!
````

**Comment by kostrzewa** [comment]
I don't understand where that is even being called, however..

**Comment by kostrzewa** [comment]
Ah, I see, in legacy_functions.R

**Comment by kostrzewa** [comment]
Okay, so after adding the transpose there and accessing the ```res``` list element, one still has to deal with the fact that the effective mass is not defined on some time slices.

```
    cat(sprintf("Time t=%d\n", t)) 
    print(str(mass))
    cat("\n\n")
```

results in

```
Time t=48
List of 18
 $ value     : num 0.103
 $ dvalue    : logi NA
 $ ddvalue   : logi NA
 $ tauint    : logi NA
 $ dtauint   : logi NA
 $ Wopt      : logi NA
 $ Wmax      : logi NA
 $ tauintofW : logi NA
 $ dtauintofW: logi NA
 $ Qval      : logi NA
 $ S         : num 1.5
 $ fgrad     : num [1:2, 1] NA NA
 $ N         : int 1158
 $ R         : int 1
 $ nrep      : num 1158
 $ data      : num [1:1158, 1:2] 0.0274 0.0216 0.0213 0.0298 0.0265 ...
 $ Gamma     :List of 1
  ..$ : num NA
 $ primary   : num 0
 - attr(*, "class")= chr [1:2] "uwerr" "list"
NULL


Error in result$mass[i] <- mass$res$value[1] : 
  replacement has length zero
```

**Comment by kostrzewa** [comment]
If I had the time I would rewrite the entire online measurements analysis so we could avoid having to deal with this...

**Comment by kostrzewa** [comment]
There also seems to be something wrong there, why does it have a value?

**Comment by kostrzewa** [comment]
Ugh, now of course ```mass$res$dvalue[1]``` does not work when this is ```NA```...

File: R/pcac.R [hunk 3]
```diff
@@ -105,10 +105,10 @@ pcac <- function(psfilename, apfilename, pafilename, from=3, to=3, fit=F, skip=0
...
     if(fit) {
-      mass <- uwerrderived(pcacfit, data=Z[,skip:(length(Z[1,]))], S=S, pl=debug, t, to, T2, pa)
+      mass <- uwerrderived(pcacfit, data=t(Z[,skip:(length(Z[1,]))]), S=S, pl=debug, t, to, T2, pa)
     }
     else {
-      mass <- uwerrderived(pcacsym, data=Z[,skip:(length(Z[1,]))], S=S, pl=debug, t, T2, pa)
+      mass <- uwerrderived(pcacsym, data=t(Z[,skip:(length(Z[1,]))]), S=S, pl=debug, t, T2, pa)
     }
     result$t[i] <- t
     result$mass[i] <- mass$value[1]
```
kostrzewa: ```$res``` missing

kostrzewa:
Okay, got it to work partially, see my pull-request to your repo: https://github.com/urbach/hadron/pull/52

I think it would probably be better if the return value when operating on NA were a vector of NAs, such that ```$res$value[1]``` is the correct way to access this in both cases.

Is there a reason, however, why you access it with an index? If one were to use 

```
xx <- uwerrderived_object$res$value
```

it would work in both cases.

kostrzewa:
Actually, no, even then it doesn't work...

kostrzewa:
```
> test <- rep(1, times=4)
> test
[1] 1 1 1 1
> a <- list(a_1 = NA,  a_2 = 1:10 )
> str(a)
List of 2
 $ a_1: logi NA
 $ a_2: int [1:10] 1 2 3 4 5 6 7 8 9 10
> test[2] <- a$a_1[1]
> test
[1]  1 NA  1  1
```

So all of this should work,  but doesn't...

R/UWerr.R
```diff
@@ -240,98 +239,110 @@ uwerrderived <- function(f, data, nrep, S=1.5, pl=FALSE, ...) {
...
-  if(R>1) {
-    chisqr <- sum((Fbr-Fb)^2*nrep)/CFbbopt
-    Qval <- 1-pgamma(chisqr/2, (R-1)/2)
-  }
-  else {
-    Qval <- NULL
-  }
+  res <- list(res=res, Wopt=Wopt, Wmax=Wmax,
+              tauintofW=tauintofW, dtauintofW=dtauintofW,
+              S=S, fgrad=fgrad, datamean=abb,
```
kostrzewa: would it not be more consistent to either have all determined quantities (value, dvalue, ddvalue, tauint, Wopt) which have the dimension ```length(Fbb)``` to either be individual elements of the result list or, conversely, all be bundled into the ```res``` sublist?

kostrzewa: ping?

kostrzewa: @urbach 
Just to clarify what I would like to do. Currently the return values of uwerrprimary and uwerrderived have different structures. In particular, the latter contains a ```res``` list element which needs to be accessed. I would like to modify the return value of uwerrderived to contain ```value, dvalue, ddvalue, tauint, ddtauint, Qval``` as vectors (and not part of ```res```. This way, uwerrprimary and uwerrderived have consistent return values, up to the fact that for the latter, these are vectors rather than single numbers.

kostrzewa: See https://github.com/urbach/hadron/pull/52

kostrzewa:
I'm afraid I forgot a few instances of the ```res``` list element, see https://github.com/urbach/hadron/pull/54

PR #33, opened 2018-05-25
add guards for symmetrisation flag to matrixfit and fit.effectivemass

kostrzewa:
Fix errors due to missing ```symmetrised``` member in cf class.

### DIFF
--- a/R/effectivemass.R
+++ b/R/effectivemass.R
@@ -148,7 +148,7 @@ fit.effectivemass <- function(cf, t1, t2, useCov=FALSE, replace.na=TRUE, boot.fi
     stop("t1 and t2 must be specified! Aborting...!\n")
   }
   tmax <- cf$Time/2
-  if( "symmetrised" %in% names(cf) ) {
+  if( "symmetrised" %in% names(cf$cf) ) {
     if(!cf$cf$symmetrised) {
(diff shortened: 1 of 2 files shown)


kostrzewa:
@martin-ueding This should fix the issue that you saw. Sorry about that!

R/effectivemass.R
```diff
@@ -148,7 +148,7 @@ fit.effectivemass <- function(cf, t1, t2, useCov=FALSE, replace.na=TRUE, boot.fi
     stop("t1 and t2 must be specified! Aborting...!\n")
   }
   tmax <- cf$Time/2
-  if(!is.null(cf$cf$symmetrised)) {
+  if( "symmetrised" %in% names(cf) ) {
```
martin-ueding: Before it was `cf$cf$symmetrised`, now it is `names(cf)`. I would think that we need `names(cf$cf)` then.

kostrzewa: Yup.

kostrzewa: Even better would be to not name this object "cf" :)

kostrzewa:
should be fixed now

kostrzewa:
You've probably modified this in the process of updating the cf class. Should I close this then?

PR #35, opened 2018-05-30
Translating Code from R to C

j-ostmeyer:


kostrzewa:
Works fine and seems to agree with the original implementation (at least visually)

PR #39, opened 2018-06-08
Analysing not every time slice

j-ostmeyer:
For some effective mass fits it can be helpful not to use all time slices because a reduction of the data might increase the stability of the inversion of the covariance matrix.

### DIFF
--- a/R/effectivemass.R
+++ b/R/effectivemass.R
@@ -140,7 +140,7 @@ bootstrap.effectivemass <- function(cf, boot.R, boot.l, seed=12345, type="solve"
   return(ret)
 }
 
-fit.effectivemass <- function(cf, t1, t2, useCov=FALSE, replace.na=TRUE, boot.fit=TRUE, autoproceed=FALSE) {
+fit.effectivemass <- function(cf, t1, t2, useCov=FALSE, replace.na=TRUE, boot.fit=TRUE, autoproceed=FALSE, every) {
   if(missing(cf) || !any(class(cf) == "effectivemass" )) {
     stop("cf is missing or must be of class \"effectivemass\"! Aborting...!\n")
   }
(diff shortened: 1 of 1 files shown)


martin-ueding:
Then let us add a `stopifnot(cf$nrObs == 1)` at the beginning of the function such that this will fail loud and clear.

kostrzewa:
It's not quite so simple. In some cases, you might actually want to extract a single number from plateau fits to multiple effective masses. This would be the case, for example, when dealing with smeared-smeard, local-local, smeared-local and local-smeared correlation functions, the effective masses of which can be fitted simultaneously in the same region if one so desires. Similarly, when one has interpolating operators of different types for some state, one might want to try to fit (with full correlation taken into account) a single number to all plateaus at the same time.

martin-ueding:
As I understand it, this is not a contradiction. If it does not work *now*, we should add the `stopifnot` to signal that this particular case is broken. It can then be subsequently fixed and the `stopifnot` removed.

kostrzewa:
This particular usecase works just fine, the more general one, where a separate fit for each observable or some subset of observables should be done and stored in some smart way is broken. If there are multiple observables in what was passed to the function, it will happily fit them simultaneously even if it shouldn't...

kostrzewa:
As a user one thus currently must take care to feed in only data which should actually be fitted simultaneously.

urbach:
> As a user one thus currently must take care to feed in only data which should actually be fitted simultaneously.

I'm not sure whether I ever indented this function to be used for more
than one obs...

--
Carsten Urbach
e-mail: curbach@gmx.de
        urbach@hiskp.uni-bonn.de
Fon   : +49 (0)228 73 2379
skype : carsten.urbach

PR #40, opened 2018-06-08
tex.catwitherror formatting fix for abs(err) > abs(x) 

kostrzewa:
hopefully fix the formatting of errors which look ambiguous either because their size exceeds the value itself or because the value is shown with N digits while the error has N+x digits


kostrzewa:
I'm still not happy with the result, I think it might take some more tweaking..

martin-ueding:
Does this do what you want?

https://github.com/martin-ueding/format-value-error

kostrzewa:
Not sure, what would it do for x=5.25, dx=2.50 ?

kostrzewa:
or more extremely, x=0.21, dx=10.21

kostrzewa:
argh, crap, now I changed the commit info for Johann's commit... bad rebase, bad!

kostrzewa:
There are still odd behaviour for things like

x=175.3, dx=23.2

which would be shown as

175.3(232)

rather unsatisfactory..

kostrzewa:
> Does this do what you want? https://github.com/martin-ueding/format-value-error

@martin-ueding out of curiosity, did you check for the particular cases that I mentioned?

x=175.3, dx=23.2
x=5.25, dx=2.50 
x=0.21, dx=10.21

It would be great, of course, if we didn't have to think about this futher...

martin-ueding:
These cases would look like this:

```
> format.val.err(175.3, 23.2, error.digits = 3)
[1] "175.3 +- 23.2"
> format.val.err(5.25, 2.50, error.digits = 3)
[1] "5.25 +- 2.50"
> format.val.err(0.21, 10.21, error.digits = 4)
[1] "0.21 +- 10.21"
```

The idea is that you can just plug in value and error and it will figure out how many significant digits to print. The `+-` syntax can be understand by the LaTeX package `siunitx` and depending on the settings it formats it as `175.3 ± 23.2` or `175.3(232)`.

But I think that you want the version in parentheses directly?

urbach:
> These cases would look like this:
> 
> ```
> > format.val.err(175.3, 23.2, error.digits = 3)
> [1] "175.3 +- 23.2"
> > format.val.err(5.25, 2.50, error.digits = 3)
> [1] "5.25 +- 2.50"
> > format.val.err(0.21, 10.21, error.digits = 4)
> [1] "0.21 +- 10.21"
> ```
> 
> The idea is that you can just plug in value and error and it will figure out how many significant digits to print. The `+-` syntax can be understand by the LaTeX package `siunitx` and depending on the settings it formats it as `175.3 ± 23.2` or `175.3(232)`.
> 
> But I think that you want the version in parentheses directly?

what is the standard way to wrote this 175.3 +- 23.2 with parentheses?

--
Carsten Urbach
e-mail: curbach@gmx.de
        urbach@hiskp.uni-bonn.de
Fon   : +49 (0)228 73 2379
skype : carsten.urbach

kostrzewa:
> @urbach wrote: what is the standard way to wrote this 175.3 +- 23.2 with parentheses?

I think one would like to get 

175.3(23.2)

as output in this case...

> @martin-ueding wrote: But I think that you want the version in parentheses directly?

Yep, and it should not be 175.3(232) because it's not clear how to read this.

Of course, when laying out final results, it is not a big deal to adjust these things on a case by case basis as required. However, for compact automated output, I would like to add some more if conditions to ```tex.catwitherror``` to make it reliable.

martin-ueding:
Johann and I added some unit tests for `tex.catwitherror` to pin down the behavior that we want to have. When the error is exactly 0, the results are not what we would like them to be.

martin-ueding:
The failures are these:

```
test_tex_catwitherror.R:25: failure: zero_error
tex.catwitherror(0.00123, 0, digits = 4, with.dollar = FALSE) not equal to "0.001230(0)".
1/1 mismatches
x[1]: "0(0)"
y[1]: "0.001230(0)"

test_tex_catwitherror.R:26: failure: zero_error
tex.catwitherror(12.345, 0, digits = 3, with.dollar = FALSE) not equal to "12.3(0)".
1/1 mismatches
x[1]: "12(0)"
y[1]: "12.3(0)"
```

martin-ueding:
@j-ostmeyer and I think that this can be merged now, but there is a review by @urbach was requested, so this PR is still locked. As we now have unit tests for this and more cases are covered than before, I think that this can be merged.

kostrzewa:
So what happens now in the case that caused me to get started on this?

kostrzewa:
I.e., what is the three-digit formatting of 172.3 with an error of 23.2 ?

kostrzewa:
There's one more case which occurs when trying to deal with error typesetting automatically which is not covered yet, but it's not a dealbreaker because the decimal point is preserved in the error:

```
tex.catwitherror(175.2, 23.3, digits = 5, with.dollar = FALSE) not equal to "175.2(23.3)".
1/1 mismatches
x[1]: "175.200(23.3)"
y[1]: "175.2(23.3)"
```

so it's still possible to interpret the number correctly. Since any publication table will be hand-checked anyway, this is not a big deal.

urbach:
there is another case not covered

    tex.catwitherror(x=1.e-35)
    Error in prettyNum(.Internal(format(x, trim, digits, nsmall, width, 3L,  : 
      invalid 'nsmall' argument

kostrzewa:
The error is in format

```
> reload("hadron"); for( i in 1:30 ){ print(i); tex.catwitherror(x=1.0*10^(-i)) }
Loading required package: hadron
[1] 1
[1] 2
[1] 3
[1] 4
[1] 5
[1] 6
[1] 7
[1] 8
[1] 9
[1] 10
[1] 11
[1] 12
[1] 13
[1] 14
[1] 15
[1] 16
[1] 17
[1] 18
[1] 19
[1] 20
[1] 21
Error in prettyNum(.Internal(format(x, trim, digits, nsmall, width, 3L,  : 
  invalid 'nsmall' argument
```

Looking at the docu of ```format```, it says:

```
nsmall: the minimum number of digits to the right of the decimal
          point in formatting real/complex numbers in non-scientific
          formats.  Allowed values are ‘0 <= nsmall <= 20’.
```

martin-ueding:
My `format_value_error` function first looks whether the `abs(log₁₀(value)) ≥ 5` and then divides out that magnitude from both value and error. In the output it then has for instance `1.23 ± 4.56) e10`. Perhaps something like this could be implemented here? Or we do not add that complexity and just require the user to bring the numbers somewhat close to 1 before formatting them. I guess in a table one would want to have a common magnitude anyway, so I'd prefer to put this burden on the user.

urbach:
> My `format_value_error` function first looks whether the `abs(log₁₀(value)) ≥ 5` and then divides out that magnitude from both value and error. In the output it then has for instance `1.23 ± 4.56) e10`. Perhaps something like this could be implemented here? Or we do not add that complexity and just require the user to bring the numbers somewhat close to 1 before formatting them. I guess in a table one would want to have a common magnitude anyway, so I'd prefer to put this burden on the user.

Well, one application for me is to print automatically e.g. the chisqr
value for a fit, which might become quite small. And I don't want to
interfere manually there...

PR #46, opened 2018-07-05
added functionality to plot a matrixfit devided by the best fit model

urbach:
- Is `plot.raw` a good name for the switch?

ToDo: once we have a two state fit this needs to be included in this plot functionality


kostrzewa:
Yes, seems to work fine.

kostrzewa:
clicked the wrong button, just wanted to comment

Issue #45, opened 2018-07-05
The sample data in hadron connected to `cf` are not up-to-date


urbach:
to the new `cf` class... E.g.

    data(samplecf)
    plot(samplecf)

leads to

    Error: any(inherits(cf, c("cf_orig", "cf_boot", "cf_jackknife"))) is not TRUE

urbach:
solved

PR #49, opened 2018-07-11
bring samplecf to new cf format

urbach:
fix for issue #45


urbach:
this should fix #45 and some more related, not reported issues and lets travis run through again. I think this could be merged straightly

PR #50, opened 2018-07-11
roxygen docu for plot.matrixfit

urbach:


kostrzewa:
you can also check in the resulting .Rd by running ```devtools::document()```, although @martin-ueding should confirm what will and what will not be overwritten by doing so.

martin-ueding:
Modern roxygen will only overwrite files that it created, so there should be nothing to worry about.

urbach:
> you can also check in the resulting .Rd by running ```devtools::document()```, although @martin-ueding should confirm what will and what will not be overwritten by doing so.

I see... I'm still not happy with this...

martin-ueding:
Okay, in what way? The checking in of built files, or the extra step in building?

kostrzewa:
Yes, neither am I, but for quickly documenting some more or less internal function and making sure that the documentation and the function signature stay up to date, it is quite convenient.

urbach:
> Yes, neither am I, but for quickly documenting some more or less internal function and making sure that the documentation and the function signature stay up to date, it is quite convenient.

yes, but to be really useful rebuild should happen automatically.

urbach:
how does it actually work: do I have to install the new package
version first and then call this docu function?

how do I include examples?

martin-ueding:
You first need to generate the new documentation, then you can install it. I usually just do this:

    R -e 'devtools::document(); install.packages(".", repo = NULL)'

For the examples, see http://r-pkgs.had.co.nz/man.html#man-functions or https://cran.r-project.org/web/packages/roxygen2/vignettes/rd.html:

> @examples provides executable R code showing how to use the function in practice. This is a very important part of the documentation because many people look at the examples before reading anything. Example code must work without errors as it is run automatically as part of R CMD check.
>
> However for the purpose of illustration, it’s often useful to include code that causes an error. \dontrun{} allows you to include code in the example that is never used. There are two other special commands. \dontshow{} is run, but not shown in the help page: this can be useful for informal tests. \donttest{} is run in examples, but not run automatically in R CMD check. This is useful if you have examples that take a long time to run. The options are summarised below.
>
> Instead of including examples directly in the documentation, you can put them in separate files and use @example path/relative/to/packge/root to insert them into the documentation.

kostrzewa:
No need to install, just run ```devtools::document()``` after starting R in the package root folder.

kostrzewa:
Or, more conveniently, as Martin said.

kostrzewa:
.Rd included in https://github.com/urbach/hadron/pull/53

martin-ueding:
> yes, but to be really useful rebuild should happen automatically.

This would be something for a build server. One does not track the generated documentation in git, then has roxygen run on the build server. The resulting files are then published as the package, perhaps on CRAN. Tracking them in git is just easier than setting up the infrastructure.

Basically Travis CI should publish the package on CRAN after a successful build.

PR #51, opened 2018-07-13
Two state fit for principal correlators

urbach:
This implements the model

    exp(-E(t-t0))(A + (1-A)exp(-DeltaE(t-t0)))

currently only for not shifted correlation matrices. 

In addition `bootstrap.nlsfit` is not able to deal with an optional derivative function of the fit function as well. Here are some bits and pieces still missing, so don't merge yet.


urbach:
if you look at the example for `bootstrap.gevp`, I used `matrixfit` and `bootstrap.nlsfit` to fit the above model to a principal correlator. The error band from `bootstrap.nlsfit` looks to me much more consistent, coming from the bootstrap samples directly. I don't understand what goes wrong in the error band computation in `plot.matrixfit`...

urbach:
btw.: this solves also issue #44 !

R/matrixfit.R
```diff
@@ -291,7 +390,14 @@ matrixfit <- function(cf, t1, t2,
     }
     par[i] <- sqrt(abs(CF$Cor[t1p1+(j-1)*Thalfp1])/0.5/exp(-par[1]*t1))
   }
-
+  cat(par, "\n")
+  ## for pcmodel we have three parameters
+  if(pcmodel) {
+    par[1] <- 0.3470281
```
urbach: uhh, this I have to fix...

kostrzewa:
> @kostrzewa the way you have implemented the errorband in plot.matrixfit appears to give results different from the one I obtain directly from the bootstrap replicates. Maybe I added some bug for the principal correlator, could you check, please?
In any ways, what speaks against using the bootstrap samples?

Yes, the errorband as it is implemented is probably not quite correct. I think I might have messed up the normalisation somewhere but never got around to fully checking this... I only ever used it is a sanity check.

The only reason for implementing it via the covariance matrix and the vector of parameter derivatives was because we supported fits and plots completely without bootstrap samples at the time (the covariance matrix was computed from the data by default). Now that we require bootstrap samples, it is much more reasonable to draw the errorband using the bootstrap standard deviation. It will also make the code much cleaner, so please feel free to throw it all out.

urbach:
> Yes, the errorband as it is implemented is probably not quite
> correct. I think I might have messed up the normalisation somewhere
> but never got around to fully checking this... I only ever used it
> is a sanity check. 
> 
> The only reason for implementing it via the covariance matrix and
> the vector of parameter derivatives was because we supported fits
> and plots completely without bootstrap samples at the time (the
> covariance matrix was computed from the data by default). Now that
> we require bootstrap samples, it is much more reasonable to draw the
> errorband using the bootstrap standard deviation. It will also make
> the code much cleaner, so please feel free to throw it all out.

okay, that makes a lot of sense. I'll put the error band from the
bootstrap sample in there.

urbach:
hmm, using the bootstrap samples does not change the picture. Something else goes wrong. I don't yet understand, what. Any suggestions?

urbach:
okay, found it... should all work now

PR #58, opened 2018-08-02
rename gamma matrix array because it conflicts with an existing name,…

kostrzewa:
… provide an accessor function with 'natural' indexing and export both. In the process, document both.


martin-ueding:
I have just checked out your branch, ran `devtools::document()` and `install.packages('.', repo = NULL)` and then started a new R session with `library(hadron)` and `??gamma` and got this output:

```
hadron::extract.obs     Extract One or More Gamma Combinations from am CMI Correlator
hadron::gm              Array of gamma structures
  Concepts: gamma
hadron::gm_mu           Accessor function for 'gm'
  Concepts: gamma
```

So it seems to work fine here.

PR #59, opened 2018-08-02
remove t0 parameter from matrixfit and extract it from the correlator…

kostrzewa:
… metadata

### DIFF
--- a/R/cf.R
+++ b/R/cf.R
@@ -166,7 +166,7 @@ cf_principal_correlator <- function (.cf = cf(), id, gevp_reference_time) {
   .cf$id <- id
   .cf$gevp_reference_time <- gevp_reference_time
 
-  class(.cf) <- append(class(.cf), 'cf_principal_correlators')
+  class(.cf) <- append(class(.cf), 'cf_principal_correlator')
   return (.cf)
 }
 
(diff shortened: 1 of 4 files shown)


kostrzewa:
@martin-ueding can you test this in your analysis ?

R/cf.R
```diff
@@ -166,7 +166,7 @@ cf_principal_correlator <- function (.cf = cf(), id, gevp_reference_time) {
   .cf$id <- id
   .cf$gevp_reference_time <- gevp_reference_time
 
-  class(.cf) <- append(class(.cf), 'cf_principal_correlators')
```
kostrzewa: good catch @urbach

kostrzewa: @martin-ueding do you agree with this change? since the mixin-constructor is singular, the name of the mixin should be too, right?

urbach:
> 
> 
> 
> > @@ -166,7 +166,7 @@ cf_principal_correlator <- function (.cf = cf(), id, gevp_reference_time) {
>    .cf$id <- id
>    .cf$gevp_reference_time <- gevp_reference_time
>  
> -  class(.cf) <- append(class(.cf), 'cf_principal_correlators')
> 
> good catch @urbach 

well, I catched it because the `bootstrap.gevp` example failed due
to this. Writing examples is really helpful! ;) There is a function
"run_examples" in the devtools package...

Issue #60, opened 2018-08-03
Get rid of warnings from zero length arrows


urbach:
`plotwitherror` produces warnings like

    Warning messages:
    1: In replayPlot(x) : zero-length arrow is of indeterminate angle and so skipped

heavily. Can we get rid of those?

urbach:
@kostrzewa do you know, why they come about?

kostrzewa:
Moving to ggplot2 would be an option, but I believe that we would first need to invest a significant amount of time to produce a production-quality theme. All the existing themes are (almost) completely unsuitable for physics publications, even when the tikz device is used for all text rendering. Also, I'm not sure if multiple error bars are supported, which plotwitherror can deal with in multiple ways.

kostrzewa:
If both the theme and the multi-errors can be addressed, ggplot2 does lead to a reduction in code complexity, because the interface is very clean, if, perhaps, sometimes a bit restrictive.

kostrzewa:
Despite the fact that the interface is so nice, I have a number of reservations with regard to ggplot2. It would be mandatory, for instance, to be able to place plot-internal legends. Square graphics are another important point. The "bw", "linedraw" and "classic" themes might be appropriate to use, but one would first have to see if we can produce some typical production-quality plots with them.

urbach:
> Despite the fact that the interface is so nice, I have a number of
  reservations with regard to ggplot2. It would be mandatory, for
  instance, to be able to place plot-internal legends. Square graphics
  are another important point. The "bw", "linedraw" and "classic"
  themes might be appropriate to use, but one would first have to see
  if we can produce some typical production-quality plots with them.


I'm no fan of ggplot2, maybe I just have not seen a reasonable theme
yet. Anyhow, its only warnings, maybe I'll switch them off for the
arrows. 

--
Carsten Urbach
e-mail: curbach@gmx.de
        urbach@hiskp.uni-bonn.de
Fon   : +49 (0)228 73 2379
skype : carsten.urbach
URL: http://www.carsten-urbach.eu

martin-ueding:
> Also, I'm not sure if multiple error bars are supported, which plotwitherror can deal with in multiple ways.

Do you mean something like this?

![multi-err](https://user-images.githubusercontent.com/976924/43647571-4c7c062c-9738-11e8-9afe-04eeb3865187.png)


```r
library(ggplot2)

df <- data.frame(x = 1:10,
                 y_val = runif(10, 1, 10),
[...]

PR #61, opened 2018-08-03
Remove warnings from zero length arrows

urbach:
but it doesn't seem to fix everything. Currently I don't exactly know, where the warnings come from...


kostrzewa:
Wouldn't it be nicer to write a wrapper function for "arrows" which does the option calls?

urbach:
well, might be...

martin-ueding:
PR #68 is another attempt to fix Issue #60. I think that is is easier because one does not have to fiddle with the warning levels.

PR #62, opened 2018-08-06
force absolute value for energies in matrixfit with pc-model

urbach:


kostrzewa:
Wouldn't it be more robust to restrict the opmizer to obtain E_2 > E_1 > 0 ?

urbach:
for `optim` bounds are implemented only for ‘"L-BFGS-B"’ and ‘"Brent"’, we are using BFGS because I found it to be most stable.
For `nls.lm` one can give upper and lower bounds. 

Why would it be more robust?

kostrzewa:
> Why would it be more robust?

you're probably right, since we have one difference and one value it should be fine

introducing a simple constraint on the amplitude might be helpful, however, to avoid those cases we saw where A was zero.

urbach:
> > Why would it be more robust?
> 
> you're probably right, since we have one difference and one value it should be fine
> 
> introducing a simple constraint on the amplitude might be helpful, however, to avoid those cases we saw where A was zero.

but in principle A could be zero, couldn't it?

urbach:
> > > Why would it be more robust?
> > 
> > you're probably right, since we have one difference and one value it should be fine
> > 
> > introducing a simple constraint on the amplitude might be helpful, however, to avoid those cases we saw where A was zero.
> 
> but in principle A could be zero, couldn't it?

or at least very small

martin-ueding:
Since we have specified the gradient ourselves there should be no issues with the non-differentiability at zero, right?

An alternative would be to just let the fit do something and then exchange the fitted masses if the mass difference is negative. I have no feeling how that compares to the current variant with the absolute value.

From looking at the code I believe this is fine, I will try out the fits tomorrow.

R/matrixfit.R
```diff
@@ -474,7 +479,9 @@ matrixfit <- function(cf, t1, t2,
   if(is.null(cf$cf)) {
     N <- cf$N
   }
-
+  if(pcmodel) {
+    opt.tsboot[,c(1:2)] <- abs(opt.tsboot[,c(1:2)])
```
martin-ueding: The dimension of `opt.tsboot` is 4 × `boot.R`. Therefore this just changes the sign of all parameters (mass, mass difference, amplitude, chi²) for the first two bootstrap samples. This instead just be

```r
opt.tsboot[c(1:2), ] <- abs(opt.tsboot[c(1:2), ])
```

kostrzewa:
@martin-ueding Can you fork @urbach's branch here and make the changes? It would be good to be able to advance in our understanding of this issue until Monday...

martin-ueding:
I have tested before and after with some particular bad fit interval. The result seems that the mass difference is now non-negative. Before it looked like this:

auswahl_005

And with the latest changes it looks better:

screenshot_20180809_152904

The fit itself is not to be taken seriously. Just for reference, it is the following:

screenshot_20180809_152856

Therefore I think that this change improves the fits and it should be merged.

kostrzewa:
Excellent

PR #63, opened 2018-08-09
Changed fit.effectivemass: optim -> weighted mean

j-ostmeyer:
This gives considerable speedup without loss of accuracy (probably the accuracy will even rise). The speedup is especially large for the uncorrelated fit.


martin-ueding:
Johann has tested this with his data and uncorrelated fits, I just tried it with a correlated fit and `samplecf`. At an effective mass of 0.2 we have absolute deviations at 1e-10, so this should give the same result as before.

PR #65, opened 2018-08-17
cf monoid

martin-ueding:
The various mixins that we have for the `cf`-class group several fields of the object. The `cf0` field still is not part of a mixin and is treated inconsistently throughout the various functions.

We had a discussion about allowing an empty `cf` object such that one can do this:

```r
corr <- cf()
for (elem in corr_list) {
    corr <- c.cf(corr, elem)
}
```

At the time I argued that this should rather be done with `do.call(c.cf, corr_list)` such that an empty `cf` object could be avoided. We do have this empty `cf` now, and it works for `c.cf`, but it does not work for `+.cf`.

Additionally we have this `cf0` field which is a strange beast. It contains the average over the original measurements. This can be computed from `cf` with `apply(mean, 2, cf)`. When one solves the GEVP on the bootstrap samples, there is no `cf` field left, we only have the bootstrap samples and the transformed `cf0`. So the `cf0` field is neither contained in `cf_orig` nor `cf_boot` nor `cf_jackknife` *exclusively*. I have the feeling that we need another mixin `cf_mean` in order to reflect that we have the means (or central values?).

A bit more abstractly we want that all `cf` objects are monoids with respect to certain binary operations, most importantly `+` and `c`. The above `for`-loop is a left fold (`foldl` in Haskell, `Reduce` in R). Both operations are a group and the `cf` class needs to be a group with a neutral element with respect to the operation.

We have the neutral element now, it is just `cf()`. And `c.cf` honors it, so we have the monoid (group without inverse) axioms:

- `c.cf(cf(), corr) == corr`
- `c.cf(corr, cf()) == corr`
- `c.cf(c.cf(corr1, corr2), corr3) == c.cf(corr1, c.cf(corr2, corr3))`

With `+.cf` and `add.cf` we do not have this, the neutral element would just give an error right now. So instead I need to do the following:

```r
corr_sum <- corr_list[[1]]
if (length(corr_list) > 1) {
    for (i in 2:length(corr_list)) {
        corr_sum <- corr_sum + corr_list[[i]]
    }
}
```

I would prefer if I can write this as a fold as I find that much more clear on intent:

```r
Reduce(`+.cf`, corr_list, cf())
```

The `cf0` field makes things more complicated. I just had a discussion with Johann and Markus and our conclusion is the following: There are certain operations that make sense on the original data (`+.cf`, `c.cf`, `addstat.cf`) and other that only make sense on bootstrapped data (`plot.cf`, `matrixfit`, …). Therefore the `cf0` should be part of the `cf_boot` mixin and `cf_orig` shall only have the `cf` and `icf` fields.

This makes it much easier to make functions like `c.cf` consistent with monoid axioms as one simply does not have to worry about `cf0`, it will be just dropped.

This pull request is not finished yet. Since we often have pull requests that are not finished yet, I'll add a tag for these situations such that we do not have to put “do not merge” at the top.


martin-ueding:
This is the old `c.cf`:

```r
c.cf <- function(...) {
  fcall <- list(...)

  # In case there is only one element, we do not need to do anything.
  if(length(fcall) == 1) {
    return(eval(fcall[[1]]))
  }

  # All arguments must be of type `cf`. Since it seems to be a common pattern
  # to concatenate stuff to an empty `cf` object, this must be supported as
  # well. We will just ignore these empty objects and concatenate the ones with
  # data.
  stopifnot(all(sapply(fcall, function (x) inherits(x, 'cf'))))
  has_data <- sapply(fcall, function (x) inherits(x, 'cf_orig'))
  fcall <- fcall[has_data]

  cf <- fcall[[1]]
  Time <- cf$Time
  cf$nrObs <- 0
  cf$sTypes <- 0
  N <- dim(cf$cf)[1]
  for (i in 1:length(fcall)) {
    if (fcall[[i]]$Time != Time) {
      stop("Times must agree for different objects of type cf\n Aborting\n")
    }
    if (dim(fcall[[i]]$cf)[1] != N) {
      stop("Number of measurements must agree for different objects of type cf\n Aborting\n")
    }
    cf$nrObs <- cf$nrObs + fcall[[i]]$nrObs
    cf$sTypes <- cf$sTypes + fcall[[i]]$sTypes
  }
  if (1 < length(fcall)) {
    for (i in 2:length(fcall)) {
      cf$cf <- cbind(cf$cf, fcall[[i]]$cf)
      cf$icf <- cbind(cf$icf, fcall[[i]]$icf)
      cf$cf0 <- c(cf$cf0, fcall[[i]]$cf0)
    }
  }
  cf <- invalidate.samples.cf(cf)

  return (invisible(cf))
}
```

And this is the new version with a binary operator abiding the monoid axioms and the general `c.cf` implemented using a fold:

```r
c.cf <- function (...) {
  rval <- Reduce(concat.cf, list(...), cf())
  return (invisible(rval))
}

#' Concatenates two correlation function objects
concat.cf <- function (left, right) {
  stopifnot(inherits(left, 'cf'))
  stopifnot(inherits(right, 'cf'))

  if (inherits(left, 'cf_boot') || inherits(left, 'cf_jackknife')
      || inherits(right, 'cf_boot') || inherits(right, 'cf_jackknife')) {
    warning('At least one argument of concat.cf has bootstrap/jacknife samples which cannot be concatenated. The samples will be discarded. If it does not have original data, it will be treated like an empty cf object.')
  }

  # In case that one of them does not contain data, the other one is the
  # result. This satisfies the neutral element axiom of a monoid.
  if (!inherits(left, 'cf_orig')) {
    return (right)
  }
  if (!inherits(right, 'cf_orig')) {
    return (left)
  }

  stopifnot(inherits(left, 'cf_meta'))
  stopifnot(inherits(right, 'cf_meta'))

  # At this point both `cf` objects given here have original data, therefore we
  # need to concatenate them.

  # A few checks for compatability.
  stopifnot(left$Time == right$Time)
  stopifnot(nrow(left$cf) == nrow(right$cf))
  stopifnot(left$symmetrised == right$symmetrised)

  rval <- cf_meta(nrObs = left$nrObs + right$nrObs,
                  Time = left$Time,
                  nrStypes = left$nrStypes, right$nrStypes,
                  symmetrised = left$symmetrised)
  rval <- cf_orig(.cf = rval,
                  cf = cbind(left$cf, right$cf),
                  icf = cbind(left$icf, right$icf))
  return (invisible(rval))
}
```

kostrzewa:
sTypes is not an attribute which should be incremented. Either all correlation functions under consideration have this many different smearign types, or none do.

martin-ueding:
I see, this is now changed. Is this ready to be merged then?

R/cf.R
```diff
@@ -474,10 +475,19 @@ avg.cbt.cf <- function(cf){
 
 ## this is intended for instance for adding diconnected diagrams to connected ones
 add.cf <- function(cf1, cf2, a=1.0, b=1.0) {
+  stopifnot(inherits(cf1, 'cf'))
+  stopifnot(inherits(cf2, 'cf'))
+
+  if (!inherits(cf1, 'cf_orig')) {
```
kostrzewa: this makes no sense to me, this should simply fail

martin-ueding:
If that fails one does not have a neutral element for the addition operation. That is the goal of this pull request to either write it as a fold operation or if preferred with a loop like this:

corr <- cf()
for (...) corr <- corr + other

Without this logic the empty cf would be less useful.

kostrzewa:
But you never want to execute arithmetic operations on an empty cf and a nonenpty cf. The empty cf is not meant as an axiomatic neutral element for this operation, it is there in order to be able build cf objects bit by bit.

The only place where it makes sense to operate on an empty cf object is in ```addStat.cf```, where either the metadata for "lhs" and "rhs" have to match exactly, or one of them is empty, such that the nonempty one is the result.

As for concatenation, this should only be possible for correlation functions with matching statistics, such that concantenation produces a statistically compatible cf object with a larger number of observables.

kostrzewa:
> But you never want to execute arithmetic operations on an empty cf and a nonenpty cf. The empty cf is not meant as an axiomatic neutral element for this operation, it is there in order to be able build cf objects bit by bit.

The reason I'm opposed to allowing arithmetic operations on empty cf objects is that one may accidentally create one in an analysis (assuming wrongly that it actually contains data). In the subsequent steps of the analysis, one might then perform arithmetic operations, which go through fine even though there is a clear programming error (from the point of view of physical logic) which could be caught at this stage.

martin-ueding:
So the addStat.cf is fine and you disagree with the need of a neutral element for addition and concatenation if I understood correctly.

Before I ever touched c.cf it already supported concatenation with an empty cf object and I remember that this was a requirement. For the GEVP I need to concatenate a list of cf objects into one. The proposed c.cf appears to be easier to understand. I do not think that there is more functionality, except when one has ill-conditioned cf objects before.

For the addition I had the case for the sigma that I had to assemble the correlation function from the diagrams. To me it now seems straightforward to map the multiplication with the factors over them and then do a fold with the addition operation.

Your point about accidentally creating empty objects is indeed a problem, that will lead to nasty bug searches. Emitting warnings is no good as then normal fold operations would give a warning. 

And I would like to apply fold operations using the binary operations. But the need for a neutral element comes just from the R "Reduce" function. In principle we could have a fold without a neutral element. So what about this: We write our own fold function like this:

foldr1 <- function (f, xs) {
  l <- length(xs)
  stopifnot(l > 0)
  if (l == 1) return (xs[[1]])
  else return (Reduce(f, xs[2:l], xs[[1]]))
}

The binary functions are changed such that they only operate on full data, everything else in a hard error. This would still allow all the functional style that I would like to have but eliminates the empty cf problem.

However, that would also break the non-functional style of creating an empty cf and then adding data to it.

So maybe allow for neutral element in some (just addStat.cf?) but not all of the operations?

urbach:
But isn't the neutral element for additions different from the  neutral element `c.cf()`? For additions you could create a `zero.cf(T, ...)`, which of course, needs proper time extent etc. but only zeros as data!?

kostrzewa:
> But isn't the neutral element for additions different from the neutral element c.cf()? For additions you could create a zero.cf(T, ...), which of course, needs proper time extent etc. but only zeros as data!?

Yes, exactly. We discussed this and have come to the conclusion that arithmetic operations should definitely error out when the user attempts to combine incompatible data (including attempting to use an empty cf with the arithmetic overloads).

On the other hand, concatenating of course should allow this with some special behaviour for cf objects without original data, where it might be desirable to also use concatenation, but only after checking that the resampling parameters are consistent (ideally one would also have a record of the configuration indices, but that's beyond the scope of this PR, I would say).

martin-ueding:
I have now a hard failure in the arithmetic operations and a monoid behavior in `c.cf` and `addStat.cf`. Is that how we want it?

The robustness in `c.cf` and `addStat.cf` is increased a bit by requiring that the empty `cf` object really is completely empty (`is.null(names(.cf))`) and that the classes match the one from the `cf()` constructor.

This way both the procedural and functional style for building up a GEVP or large statistics work robust. Also there is a `foldr1` function that can be used with the arithmetic operations if one wants to. There is no danger of accidentally using an empty object with arithmetics.

R/cf.R
```diff
@@ -400,10 +405,19 @@ addConfIndex2cf <- function(cf, conf.index) {
 }
 
 addStat.cf <- function(cf1, cf2) {
+  stopifnot(inherits(cf1, 'cf'))
+  stopifnot(inherits(cf2, 'cf'))
+
+  if (is_empty.cf(cf1)) {
+    return (invisible(cf2))
+  }
+  if (is_empty.cf(cf2, 'cf_orig')) {
```
kostrzewa: why two arguments?

martin-ueding: That was just negligence, it is fixed now.

martin-ueding:
Is there something that needs to be resolved before I can merge this pull request?

Issue #66, opened 2018-08-17
symmetrise.cf is not used in readtextcf


martin-ueding:
In `readtextcf` I noticed that symmetrization is done “by hand” and I thought that we do not have a function for symmetrization. Just now I have found `symmetrise.cf`; it should be used in the IO functions.

martin-ueding:
See #67.

PR #67, opened 2018-08-17
Use symmetrize.cf in reading functions

martin-ueding:
See #66.


R/readutils.R
```diff
@@ -412,10 +395,16 @@ readbinarycf <- function(files,
     }
   }
 
-  cf <- cf_meta(nrObs = 1, Time=T, nrStypes = 1, symmetrised = symmetrise)
-  cf <- cf_orig(cf, cf = t(Re(Cf)), icf = t(Cf))
+  ret <- cf_meta(nrObs = 1, Time=T, nrStypes = 1, symmetrised = symmetrise)
+  ret <- cf_orig(ret, cf = t(Re(Cf)), icf = t(Cf))
```
kostrzewa: ```Im``` missing

martin-ueding:
It was missing in two cases, `readtextcf` seemed to be correct. Is this okay to merge now?

R/readutils.R
```diff
@@ -470,7 +451,11 @@ readbinarysamples <- function(files, T=48, nosamples=2, endian="little",
   ret <- list()
   for (i in 1:nosamples) {
     ret[[i]] <- cf_meta(nrObs = 1, Time=T, nrStypes = 1, symmetrised = symmetrise)
```
kostrzewa: this is actually currently broken because ```symmetrise``` does not exist...

kostrzewa:
The whole set of reading routines needs some work because some things have become inconsistent. Most notably how smearing and observables are treated...

R/readutils.R
```diff
@@ -458,9 +442,6 @@ readbinarysamples <- function(files, T=48, nosamples=2, endian="little",
         } else {
           tmp2 <- apply(X=tmp[,1:i],MARGIN=1,FUN=mean)
         }
-        # average over +- t
-        tmp2[i1] <- 0.5 * ( tmp2[i1] + sign * tmp2[i2] )
-        Cf[[i]] <- cbind(Cf[[i]],tmp2[1:(T/2+1)])
```
kostrzewa: you broke functionality here

R/readutils.R
kostrzewa: ping?

kostrzewa:
@martin-ueding is this abandoned?

martin-ueding:
Sorry for not getting back to this. I have just looked at it again and realize that indeed this breaks things. In the constructor I now use `symmetrize = TRUE`, so that should be fine. And for the actual data I now append the full correlation function and then symmetrize them at the end with the new function.

Is there some sample data where I could test this with?

kostrzewa:
There's also still the problem with the ```readbinarysamples``` function.

I've added example data to the package in https://github.com/hiskp-lqcd/hadron_example_data

martin-ueding:
I have just looked into this sample data repository but I only found text files containing data. If the error is in the `readbinarysamples` function, I would need binary data, right?

martin-ueding:
I have tested the `readbinarysamples` function with the test data, the shape of the output looks sensible now.

PR #68, opened 2018-08-17
Do not draw zero length arrows

martin-ueding:
See #60.

The logic for checking zero length arrows was in place already, so I have just skipped drawing of a zero length arrow. In the example one can see that the warning is not emitted any more.

### DIFF
--- a/R/plotutils.R
+++ b/R/plotutils.R
@@ -188,7 +188,7 @@ plotwitherror <- function(x, y, dy, ylim, dx, xlim, mdx, mdy, errsum.method="lin
           start <- y[rw]+cumul.err[rw,(level-1)]
           end <- y[rw]+cumul.err[rw,level]
 
-          if (start != end) {
+          if (!is.na(start) && !is.na(end) && start != end) {
             arrows(x[rw], start, x[rw], end, length=arwhd.len, angle=90, code=2, col=clr)
             arwhd.len <- arwhd.len + 0.01
           }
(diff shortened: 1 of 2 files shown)


martin-ueding:
Do you understand why it fails to build on Travis CI? The problem seems to be there in line 1931:

```
> plot(pc1.effectivemass)
Execution halted
```

Why would that happen?

urbach:
Martin, did you try `example(bootstrap.effectivemass)` yourselves?

urbach:
or is it `example(bootstrap.gevp)` here?

martin-ueding:
It is the example of `bootstrap.gevp`. One can run the examples without installing the package using `devtools::dev_example('bootstrap.gevp')` by the way.

The error that occurs within `plotwitherror` because `begin` and `end` are both `NA`:

```r
        for(level in rng){
          start <- y[rw]+cumul.err[rw,(level-1)]
          end <- y[rw]+cumul.err[rw,level]

          if (start != end) {
            arrows(x[rw], start, x[rw], end, length=arwhd.len, angle=90, code=2, col=clr)
            arwhd.len <- arwhd.len + 0.01
          }
        } 
```

The example works just fine using the `master` version, so I must have broken this. I'll see to it.

martin-ueding:
I have added this check as the main addition in this pull request. Now I have just extended it to make sure that neither `start` nor `end` are `NA`. That has fixed the example locally, let us see how it goes on Travis CI.

kostrzewa:
very nice

Issue #17, opened 2017-08-16
uwerrprimary does not return the error on Gamma


martin-ueding:
The `uwerrprimary` function returns a `list` with lots of variables, but the error on Gamma is missing. It is computed as needed in the `plot.uwerr` function. I'd like to extract the data and plot with some different package, therefore I'd need this data directly.

I have looked into the code and in `uwerrprimary` there is a `GammaFbb` which seems different from the `GammaFbb` that is used in `plot.uwerrprimary`.

Could the error on Gamma be also included in the return value of this function?

*Update*: It seems that I cannot use the `gammaerror` function in my own script because that does not get exported into the main namespace, it seems. I suppose this is a function with file-scope only, then.

*Update*: I managed to import the `uw.R` source file and then just copied the needed code from the plotting routine. So I have the data that I want, so this issue can wait.

martin-ueding:
It's not urgent, I got the data that I need, so everything is fine.

urbach:
still needed?

urbach:
And, `prot.uwerr` does return the autocorrelation function with error bars.

PR #70, opened 2018-08-17
Let uwerrprimary return error on Gamma

martin-ueding:
See Issue #17.

### DIFF
--- a/R/UWerr.R
+++ b/R/UWerr.R
@@ -124,6 +124,7 @@ uwerrprimary <- function(data, nrep, S=1.5, pl=FALSE) {
     stop("Gamma pathological: error^2 <0")
   }
   GammaFbb <- GammaFbb+CFbbopt/N            #Correct for bias
+  dGamma <- gammaerror(Gamma=GammaFbb, N=N , W=Wmax, Lambda=100)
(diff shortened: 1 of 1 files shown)


kostrzewa:
This looks fine, but I don't know why gammaerror depends on Lambda and why ```gamma.err[1]``` is hard-coded

kostrzewa:
Lambda is completely ignored (in the formula in question, it is the maximum lag used to evaluate the variance of the normalised autocorrelation function). The paper states that it is not crucial what the value of Lambda is, but here it seems that W is used instead.

urbach:
okay, `Lambda` is not used indeed and should be removed!

I think using `W` should be sufficient here, shouldn't it?

kostrzewa:
I would agree, yes. I think we can use it as is. I am, however, a bit perplexed as to why the very first error is hardcoded.

urbach:
> I would agree, yes. I think we can use it as is. I am, however, a bit perplexed as to why the very first error is hardcoded.

Well, acutally the error is identically zero, because the value is 1
by construction. But since 0 is not easy to draw, I put a 0.001, not
very elegantly, I agree...

kostrzewa:
I see, this is ready to go in then, I would say.

kostrzewa:
I though the first value was lag '1' rather than lag '0', my bad.

PR #74, opened 2018-09-13
New bootstrap.nlsfit

j-ostmeyer:
The new implementation of `bootstrap.nlsfit()` is tested for all eventualities that I could think of.

It has the following advantages over the original version:

- Much better runtime in case of no covariance and many data points.
- Less redundant function definitions.
- Simple parallelisation supported.
- Parametric bootstrap with covariance possible.
- Improved plot function.

The documentation is yet to be changed.


R/bootstrap.nlsfit.R
```diff
@@ -211,44 +251,24 @@ print.bootstrapfit <- function(x, digits=2, ...) {
...
-    }
-    if(missing(ylim)) {
-      my.ylim <- mylims$ylim
-    }
-    
-    ## generate empty plot
-    
-    plot(NA, xlim=my.xlim, ylim=my.ylim, xlab=xlab, ylab=ylab, ...)
+  else {
+    mylims <- plotwitherror(x=x$x, y=x$y, dy=x$dy, dx=x$dx, rep=rep, xlim=xlim, ylim=ylim, ...)
```
urbach: I did the `plotwitherror` at the end of everything on purpose, such that the data points are on top of everything else. It seems this is changed now?

R/bootstrap.nlsfit.R
```diff
@@ -264,18 +284,13 @@ plot.bootstrapfit <- function(x, ..., xlim, ylim, rep=FALSE, col.line="black", c
...
   ## plot the fitted curve on top
   lines(x=X, y=Y, col=col.line, lty=lty, lwd=lwd)
-
-  ## plot data on top of everything
-  if(x$errormodel == "yerrors") {
-    plotwitherror(x=x$x, y=x$y, dy=x$dy, rep=TRUE)
-  }
-  else {
-    plotwitherror(x=x$x, y=x$y, dy=x$dy, dx=x$dx, rep=TRUE)
-  }
```
urbach: see comment above

urbach:
The interface has changed, but the docu page didn't, did it?

R/bootstrap.nlsfit.R
```diff
@@ -1,100 +1,47 @@
 bootstrap.nlsfit <- function(fn,
                              gr = NULL,
+                             dfn = NULL,
```
urbach: what is `dfn` in contrast to `gr`?

R/bootstrap.nlsfit.R
```diff
@@ -1,100 +1,47 @@
...
-    else res <- optim(par=par, fn=fitchisqr, gr=dfitchisqr, y=y, method=c("BFGS"), fitfun=fitfun, dfitfun=dfitfun, dy=dy, x=x, ...)
-
-    return (c(res$par, res$value))
-  }
-  wrapper.optim.xy <- function(y, par, fitfun, dy, nx, ...) {
-    res <- optim(par=par, fn=fitchisqr.xy, y=y, method=c("BFGS"), fitfun=fitfun, dy=dy, nx=nx, ...)
-    return (c(res$par, res$value))
+  lm.avail <- require(minpack.lm)
+  if(parallel){
+    parallel <- require(parallel)
```
urbach: this package needs to be added to the hadron package description.

R/bootstrap.nlsfit.R
```diff
@@ -113,55 +60,146 @@ bootstrap.nlsfit <- function(fn,
...
-  }
-  else if(useCov) {
-    CovMatrix <- cov(bsamples)
-    InvCovMatrix <- try(invertCovMatrix(bsamples, boot.l=1, boot.samples=TRUE), silent=TRUE)
-    if(inherits(InvCovMatrix, "try-error")) {
-      stop("Variance-covariance matrix could not be inverted!")
+    if(missing(CovMatrix)){
+      InvCovMatrix <- try(invertCovMatrix(bsamples, boot.l=1, boot.samples=TRUE), silent=TRUE)
+      inversion.worked(InvCovMatrix)
+      dY <- chol(InvCovMatrix)
```
urbach: after our discussion, why do you need the Cholesky decomposition here?

R/bootstrap.nlsfit.R
```diff
@@ -211,44 +251,24 @@ print.bootstrapfit <- function(x, digits=2, ...) {
   summary.bootstrapfit(object=x, digits=digits, ...)
 }
 
-plot.bootstrapfit <- function(x, ..., xlim, ylim, rep=FALSE, col.line="black", col.band="gray", lty=c(1), lwd=c(1), xlab="x", ylab="y") {
+plot.bootstrapfit <- function(x, ..., xlim, ylim, rep=FALSE, col.line="black", col.band="gray", lty=c(1), lwd=c(1), xlab="x", ylab="y", samples=1000) {
```
urbach: `samples` might be a bit misleading as a name here

R/bootstrap.nlsfit.R
j-ostmeyer: Yes, I changed this for the following reasons:
-In the way it has been, there have been always two plots created when using a script and producing a pdf. The first one contained only the data, the second one the fit and the data.
-If you have a lot of data and plot it on top of the fit, you don't see the fit any more. But if you plot the data first and then the fitline you still see everything because the errorband is transparent.
-The way it is now is consistent with plot.effectivemass where the fit is plotted on top as well.
-With this function one wants to plot the results of the fit, not the data, so this is where the focus should lie.

j-ostmeyer:
> The interface has changed, but the docu page didn't, did it?

No, it did not. See my very first comment.

R/bootstrap.nlsfit.R
j-ostmeyer: gr = d(fn) / d(par)
This is the gradient in parameter space.

dfn = d(fn) / dx
This is the derivative by the canonical variable. Only needed for the gradient function with xyerrors.

R/bootstrap.nlsfit.R
j-ostmeyer: I don't really need it here.
But I need it for parametric bootstrap with covariance. And I need it in case the gradient is used.
So I simply left it here for consistency. I am aware that this is not the most performant way, but the difference should not be significant because there is the inversion anyway.

R/bootstrap.nlsfit.R
j-ostmeyer: This is the gnuplot term. But I get your point. I can change it to whatever name you prefer.

urbach:
> j-ostmeyer commented on this pull request.
> 
> 
> 
> > -      else {
> -        mylims <- plotwitherror(x=x$x, y=x$y, dy=x$dy, dx=x$dx, ...)
> -      }
> -    }
> -    if(missing(xlim)) {
> -      my.xlim <- mylims$xlim
> -    }
> -    if(missing(ylim)) {
> -      my.ylim <- mylims$ylim
> -    }
> -    
> -    ## generate empty plot
> -    
> -    plot(NA, xlim=my.xlim, ylim=my.ylim, xlab=xlab, ylab=ylab, ...)
> +  else {
> +    mylims <- plotwitherror(x=x$x, y=x$y, dy=x$dy, dx=x$dx, rep=rep, xlim=xlim, ylim=ylim, ...)
> 
> Yes, I changed this for the following reasons:
> -In the way it has been, there have been always two plots created when using a script and producing a pdf. The first one contained only the data, the second one the fit and the data.
> -If you have a lot of data and plot it on top of the fit, you don't see the fit any more. But if you plot the data first and then the fitline you still see everything because the errorband is transparent.
> -The way it is now is consistent with plot.effectivemass where the fit is plotted on top as well.
> -With this function one wants to plot the results of the fit, not the data, so this is where the focus should lie.

this we need to discuss!

urbach:
> j-ostmeyer commented on this pull request.
> 
> 
> 
> > @@ -1,100 +1,47 @@
>  bootstrap.nlsfit <- function(fn,
>                               gr = NULL,
> +                             dfn = NULL,
> 
> gr = d(fn) / d(par)
> This is the gradient in parameter space.
> 
> dfn = d(fn) / dx
> This is the derivative by the canonical variable. Only needed for the gradient function with xyerrors.

so the interface changed...!?

man/fit.cosh.Rd
```diff
@@ -0,0 +1,96 @@
+\name{fit.cosh}
```
kostrzewa: Why are you fitting a sum of cosh to an effective mass?

j-ostmeyer: I fit the correlators, but I use the effMass class to do so. This is because it is an alternative for fit.effectivemass and can be used in precisely the same way. The idea is that the fit result can be compared to the effective mass plateaus directly.

kostrzewa: Okay, but then this should also support fitting to a bootstrapped ```cf``` directly.

martin-ueding:
It feels a bit like a violation of the Single Responsibility Principle. Wouldn't it be better to have the fit work with cf_boot objects and add an effmass parameter to the plot function such that one can compare all fits (matrixfit) with an effective mass?

R/bootstrap.nlsfit.R
```diff
@@ -173,12 +216,12 @@ bootstrap.nlsfit <- function(fn,
...
 }
 
-summary.bootstrapfit <- function(object, digits=2, ...) {
+summary.bootstrapfit <- function(object, digits=2, print.correlation=TRUE, ...) {
 
   cat("bootstrap nls fit\n\n")
   cat("model", object$errormodel, "\n")
   errors <- object$se
-  values <- object$t[1, 1:(dim(object$t)[2]-1)]
+  values <- object$t0[1:(length(object$t0)-1)]
```
kostrzewa: -1?

R/cosh_nlsfit.R
```diff
@@ -0,0 +1,210 @@
...
+		stopifnot(inherits(cf, 'cf_boot'))
+		Thalf = cf$Time/2
+		use.effmass <- FALSE
+	}
+
+	stopifnot(t1 < t2)
+	stopifnot(0 <= t1)
+	
+	if(missing(par)){
+		## some initial values, probably not optimal...
```
kostrzewa: please don't use tabulators for indentation

j-ostmeyer:
From my perspective this can be merged. Are there any open issues I did not consider?

kostrzewa:
The point by @martin-ueding  bears consideration, although it is fine to go in from my point of view because technically the effective mass should be an extension of ```cf``` and not an essentially separate class. 

What about the concerns by @urbach with regard to the ordering of drawing operations?

kostrzewa:
> The point by @martin-ueding bears consideration, although it is fine to go in from my point of view because technically the effective mass should be an extension of cf and not an essentially separate class.

I'm aware of course of the complications this would entail with regard to the various plotting routines and/or summary functions, although I would say that if one plots the effective mass, one would probably also like to plot some or all of the plots showing the correlation function and fits to it.

R/bootstrap.nlsfit.R
kostrzewa: @j-ostmeyer also this should be at least commented on

j-ostmeyer:
I get the point, but this would mean a lot of changes. Does it make sense to do them in this pull request?

kostrzewa:
> You can actually choose whether you want to plot the effective mass (if you have it) or the correlator or both. There are corresponding flags in the plot function.

No no, not at all, that's a story for another week/month/year.

R/cosh_nlsfit.R
```diff
@@ -0,0 +1,210 @@
...
+fit.cosh <- function(effMass, cf, t1, t2, useCov=FALSE, m.init=0.01, par, n.cosh=2, every, ...){
+  if(missing(effMass) && missing(cf)){
+    stop("effMass or cf has to be provided!\n")
+  }
+  if(!missing(effMass)){
+    stopifnot(inherits(effMass, 'effectivemass'))
+    Thalf = effMass$Time/2
+    use.effmass <- TRUE
+  }else{
+    stopifnot(inherits(cf, 'cf_boot'))
```
urbach: shouldn't there be also `stopifnot(inherits(cf, 'cf_meta'))`?

R/bootstrap.nlsfit.R
kostrzewa: ping

martin-ueding: This is resolved, in the current state of the branch there is `parallel` in the `Suggests` of the `DESCRIPTION`.

R/cosh_nlsfit.R
martin-ueding: Indeed, as `cf$Time` is accessed one needs the `cf_meta` mixin. This has been added in the latest state of Johann's branch.

martin-ueding:
I believe that now all the things in the scope of this pull request have been resolved and that we should merge it. Any objections?

kostrzewa:
Please go ahead.

urbach:
yes, please go ahead.

> Please go ahead.
> 
> -- 
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub:
> https://github.com/HISKP-LQCD/hadron/pull/74#issuecomment-432175857
--
Carsten Urbach
e-mail: curbach@gmx.de
        urbach@hiskp.uni-bonn.de
Fon   : +49 (0)228 73 2379
skype : carsten.urbach

PR #76, opened 2018-10-05
Plotutils and gradflow

kostrzewa:
* fix an issue with logarithmic limits in plotwitherror
* rework gradient flow analysis to produce some more output, in particular to generate two estimates of t_0 and to save some summary information in a named list much like analysis_online

### DIFF
--- a/R/plotutils.R
+++ b/R/plotutils.R
@@ -9,11 +9,11 @@ compute.plotlims <- function(val, logscale, cumul.dval, cumul.mdval){
   if(logscale) {
     tmp <- tmp[ tmp > 0 ]
     tmpp <- tmpp[ tmpp > 0 ]
-  }
-  if( ( all(is.na(tmp)) && all(is.na(tmpp)) ) | ( length(tmp) == 0 & length(tmpp) == 0 ) ){
-    warning("compute.plotlims: log scale requested but there are no positive data, setting default range\n")
-    tmp <- 10^(-6)
-    tmpp <- 10^2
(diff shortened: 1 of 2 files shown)


R/plotutils.R
```diff
@@ -9,11 +9,11 @@ compute.plotlims <- function(val, logscale, cumul.dval, cumul.mdval){
   if(logscale) {
     tmp <- tmp[ tmp > 0 ]
     tmpp <- tmpp[ tmpp > 0 ]
-  }
-  if( ( all(is.na(tmp)) && all(is.na(tmpp)) ) | ( length(tmp) == 0 & length(tmpp) == 0 ) ){
-    warning("compute.plotlims: log scale requested but there are no positive data, setting default range\n")
-    tmp <- 10^(-6)
-    tmpp <- 10^2
+    if( ( all(is.na(tmp)) && all(is.na(tmpp)) ) | ( length(tmp) == 0 & length(tmpp) == 0 ) ){
```
martin-ueding: Why is this bracket clearly intended? The error message specifically states that logscale was requested, so this edge case handling is just fine in the `if`, isn't?

martin-ueding:
I have looked through the changes but I did not know or use this part of the core before, therefore I am not sure whether I can be of much help. Does it make sense for me to understand it fully and test it with some data set?

R/plotutils.R
kostrzewa: Because the default range would be used if all data were negative even if no log scale were requested. This was intended purely in case of the log scale.

kostrzewa:
Thanks, the request was more about catching obvious silly things. I've tested the code on data and it reproduces what it did before the changes.

PR #82, opened 2018-10-23
Fix jackknife

kostrzewa:


kostrzewa:
actually, I was wrong, the code as it is produces error estimates on cf which are consistent with bootstrap, some automatic transposition must take place which confused me

PR #83, opened 2018-10-23
Fix jackknife sampling and error estimation

martin-ueding:
Following our [previous discussion](https://github.com/kostrzewa/LuescherAnalysis/issues/64) these are the changes that I believe to be necessary to make the jackknife work correctly.

### DIFF
--- a/R/cf.R
+++ b/R/cf.R
@@ -352,7 +352,7 @@ jackknife.cf <- function(cf, boot.l=2) {
     ii <- c(i:(i+boot.l-1))
     ## jackknife replications of the mean
     gammai <- apply(cf$cf[-ii, ], MARGIN=2L, FUN=mean)
-    cf.jackknife$t[i, ] <- (n*cf$cf0 - (n - boot.l)*gammai)/boot.l
+    cf.jackknife$t[i, ] <- gammai
   }
   ## the jackknife error
   tmp <- apply(cf.jackknife$t, MARGIN=1L, FUN=function(x,y){(x-y)^2}, y=cf$cf0)
--- a/R/cf.R
+++ b/R/cf.R
@@ -357,7 +357,7 @@ jackknife.cf <- function(cf, boot.l=2) {
   ## the jackknife error
   tmp <- apply(cf.jackknife$t, MARGIN=1L, FUN=function(x,y){(x-y)^2}, y=cf$cf0)
   jackknife.se <- apply(tmp, MARGIN=1L,
-                           FUN=function(x, l, n, N) {sqrt( l/(n-l)/N*sum( x ) ) },
+                           FUN=function(x, l, n, N) {sqrt( (n-l)/l/N*sum( x ) ) },
                            n=n, N=N, l=boot.l)
 
   cf <- cf_jackknife(cf,


martin-ueding:
These are my comments from the other issue, which is a private repository.

---

The hadron jackknife function looks very strange:

```{r}
    # The measurements that we are going to leave out.
    ii <- c(i:(i+boot.l-1))
    ## jackknife replications of the mean
    gammai <- apply(cf$cf[-ii, ], MARGIN=2L, FUN=mean)
    cf.jackknife$t[i, ] <- (n*cf$cf0 - (n - boot.l)*gammai)/boot.l
```
So `ii` are the measurement indices that we want to leave out for the sample number `i`. Since we have `boot.l = 1` here, `ii` is just `i`. `gammai` contains the average over all data except the one that one want to leave out.

Now `cf$cf0` is the mean over all data, so `n*cf$cf0` is the sum over all data. `(n - boot.l)*gammai` then is the sum over all the data except the one we left out. This difference `n*cf$cf0 - (n - boot.l)*gammai` then is just the sum of the measurements `ii`.

This makes no sense to us because we think that `gammai` should be the sample `i`. Are we missing something? We thought that perhaps one has to set `boot.l` to `N - 1` or so in order to average over many more. But then we would only get one or two samples, not what we want either.

---

I think that the error estimate in the master branch only produces the correct results because it computes the standard error of original distribution. But I believe that the sampling and also the error estimate are incorrect, even though they produce sensible result in the current combination.

kostrzewa:
Definitions:

1) ```x_i``` is a measurement
1) ```\theta``` is an estimator ```s(X)``` on the full sample
1) ```\theta_i``` is an estimator ```s(X_[i])``` on the set of measurements ```j \neq i```, where the notation ```X_[i]``` implies leaving out the ith value.
1) ```\theta_J = 1/n \sum_i \theta_i``` is the empirical average of this estimator over the jackknife samples
1) The jacknife standard error is given by 
```\sqrt{ \frac{n-1}{n} \sum_i^n ( \theta_i - \theta_J )^2 }```

In the special case our estimator is the sample mean

6) ```\theta_i = 1/(n-1) \sum_{j \neq i} x_j``` 
7) ```\theta_J = 1/n \sum_i \theta_i = 1/n 1/(n-1) \sum_i \sum_{j \neq i} x_j``` is the jacknife estimate of the sample mean and coincides with the sample mean

For general estimators,

```\theta_{jack} = n\theta - (n-1) \theta_J``` is a bias-corrected estimator, which is what the formula in the master branch implements for the mean as soon as its averaged over the jackknife samples. This then corresponds to

```
1/n \sum_i (n \theta - (n-1)\theta_J)
= 1/n \sum_i n \theta - 1/n \sum_i (n-1) 1/n \sum_k \theta_k
= 1/n \sum_i n \theta - 1/n \sum_i (n-1) 1/n \sum_{k} 1/(n-1) \sum{j \neq k} \theta_j
```

Not quite sure how to approach the variance in this case, to be honest.

**Comment by martin-ueding** [comment]
This `\theta_J` indeed is the sample mean:



This `\theta_{jack}` as you have written it is also just the sample mean:



But that is not what I think is implemented, especially since the above is just one quantity and we want to have `n` jackknife samples with our delete-1-jackknife. What I believe to be implemented for the samples in `master` is the following, where one does not subtract `\theta_J` but `\theta_i`:



And this is just the `i`-th measurement, which I can also see by comparing the jackknife distribution with the original distribution.

Shall we discuss that in person on Thursday morning?

**Comment by urbach** [comment]
I think that my implementation of jackknife is indeed correct for the mean. However, it will not work for other estimators. So we should change to the usual way of implementing the jackknife...

**Comment by urbach** [comment]
you mean me?

**Comment by martin-ueding** [comment]
I feel addressed by @kostrzewa, I will try this out.

**Comment by kostrzewa** [comment]
Yes, I meant @martin-ueding

**Comment by martin-ueding** [comment]
The normalization is correct. From the equations that I have seen one has to divide by `(n - l)/(l * N)` where `n` is the number of measurements, `l` is the order of the jackknife and `N` is the number of samples that one has created. In the literature one often finds the binomial coefficient `n over l` for `N`. The way the jackknife is used here, one does not do any permutations possible but deletes `l` *consecutive* measurements, yielding only `N = n - l + 1` jackknife samples.

Also this can be tested numerically for different `l` and it seems to work just fine:

```r
# Generate measurements.
n <- 3000
X <- rnorm(n)

# We know what the standard error of the data is.
se <- sd(X) / sqrt(n)

# Creating bootstrap samples and estimating the error of the mean.
boot_R <- 1500
X_boot <- boot(X, function (data, indices) mean(data[indices]), boot_R)$t
se_b <- sd(X_boot)

# Do delete-l jackknife and compute the error.
l <- 40
X_jack <- sapply(l:length(X), function (i) mean(X[-c(i:(i-l+1))]))
se_j <- sqrt((n-1)/n/l * sum((X_jack - mean(X_jack))^2))

# Take a look at the errors.
se_j / se_b
se_j / se
```

The output is:

    [1] 0.8989186
    [1] 0.9166975

We see that the jackknife seems to underestimate the error a bit for large `l`. In general it seems be just statistical fluctuation, though:

error

Therefore I would conclude that the jackknife is now implemented correctly, even with `l > 1`.

kostrzewa:
Very nice, thanks checking!

Issue #81, opened 2018-10-21
bootstrap.nlsfit does not always need boot.R as a parameter


martin-ueding:
I think making `boot.R = nrow(bsamples)` the default value would work just fine in the case of bootstrap samples such that one does not have to specify this redundantly. Or only access this parameter if one does not pass `bsamples`.

kostrzewa:
Agreed, we should keep things consistent.

martin-ueding:
In principle we can also extract `dy` and `dx` from `bsamples`.

But actually I think that we are doing two things here that should be split. There should be a very simple function `parametric_bootstrap(x, dx, boot.R)` that returns a matrix. And then the `bootstrap.nlsfit` should only take bootstrap samples, so the parameters `dy`, `dx` and `boot.R` would be taken away. I would not mind adding a convenience wrapper which does the parametric bootstrap and then calls the focused `bootstrap.nlsfit`.

 Would you agree?

PR #84, opened 2018-10-29
Refactor nlsfit

martin-ueding:
I found that the parameters of `bootstrap.nlsfit` are actually three interfaces at once and made it cumbersome to use. In this pull request I have split it into three parts:

1. Two functions for parametric bootstrap, one with covariance and the over without. This works for arbitrary many variables.
2. An NLS fit function which always works on bootstrap samples. The number of parameters has been greatly reduced and I find it much more focused now. The optional parameters have been moved to the end of the parameter list. Parameters like `useCov` and `errormodel` are computed from the inputs as the presence of a covariance matrix and the number of columns of the bootstrap samples already contains all the information.
3. Convenience functions that first call the parametric bootstrap and then perform the fit.

I have not tested all possibilities, but at least the case with `yerrors` and no covariance works. This should be tested by the users, especially since the whole API has changed.

While I was at it, I have also split up the documentation into individual Roxygen comments. Documenting a family of functions which do not share parameters in a single documentation file seems hard to read for me. Now all functions are on individual pages but with a family crosslink between all of them.


martin-ueding:
This pull request addresses Issue #81.

martin-ueding:
Previously one could specify `useCov` and no covariance matrix, then the covariance matrix would be computed. This likely is something different than just passing `cov(bsamples)`?

I am a bit confused about the purpose of `dY`. It seems to hold the errors of both `y` and `x` if needed. But the line `dY <- 1.0/dY` reads like a red flag as the variable suddenly becomes the weight from now on. Is that the correct assessment? If do, I would want to introduce a new variable name.

j-ostmeyer:
In the documentation the example call to `bootstrp.nlsfit` does not work because `bsamples` are missing.

j-ostmeyer:
I agree that the dot in R is somehow misleading, but I would still appreciate consistency between the functions. I.e. the number of bootstrap samples should be called `boot.R` or `boot_R` everywhere in order to avoid confusion. Probably one should just stick to the default `boot.R` here.
This is less crucial in new functions. But if there are any other regularly used variables renamed in this pull request, the same argument would apply.

martin-ueding:
I have replaced all the underscores with periods, this way the whole hadron library stays consistent with period. This consistency is most important, although I really prefer the underscore myself.

The example of `bootstrap.nlsfit` needed to be adapted to the new API, but it somehow does not work:

```r
## Declare some data.
value <- c(0.1, 0.2, 0.3)
dvalue <- c(0.01, 0.01, 0.015)
x <- c(1, 2, 3)
dx <- c(0.1, 0.1, 0.1)
boot.R <- 1500

fn <- function (par, x) par[1] + par[2] * x

## Before we can use the fit with this data, we need to create bootstrap
## samples. We do not want to use the correlation matrix here. Note that you
## can simply use the parametric.nlsfit function as a convenient wrapper of
## the two steps.
bsamples <- parametric.bootstrap(boot.R, c(value, x), c(dvalue, dx))
head(bsamples)
#>            [,1]      [,2]      [,3]      [,4]     [,5]     [,6]
#> [1,] 0.08599956 0.1942148 0.3168092 1.0685715 1.960543 3.077450
#> [2,] 0.10255317 0.1983089 0.3059985 0.8425810 1.784644 2.700162
#> [3,] 0.07562736 0.1808077 0.2852321 0.8960797 1.912874 3.079438
#> [4,] 0.09994429 0.1846573 0.2924616 1.1062893 1.993869 3.105427
#> [5,] 0.10621553 0.1888524 0.3148072 0.8720619 1.819544 2.992919
#> [6,] 0.11148412 0.2159781 0.3328722 0.8234740 2.112201 2.934894

fit.result <- bootstrap.nlsfit(fn, c(1, 1), value, x, bsamples)
#> Loading required package: minpack.lm
summary(fit.result)
#> bootstrap nls fit
#> 
#> model xyerrors 
#> Error in while (round(10^N * err) < threshold) {    N <- N + 1}: Fehlender Wert, wo TRUE/FALSE nötig ist
```

When using the convenience function, it works just fine:

```r

## Declare some data.
value <- c(0.1, 0.2, 0.3)
dvalue <- c(0.01, 0.01, 0.015)
x <- c(1, 2, 3)
dx <- c(0.1, 0.1, 0.1)
boot.R <- 1500

fn <- function (par, x) par[1] + par[2] * x

fit.result <- parametric.nlsfit(fn, c(1, 1), boot.R, value, dvalue, x, dx)
#> Loading required package: minpack.lm
summary(fit.result)
#> bootstrap nls fit
#> 
#> model xyerrors 
#>     best fit parameters with errors, bootstrap bias and 68% confidence interval
#> 
#>         par     bias   ci16  ci84
#> 1 0.000(23)   0.0012 -0.024 0.022
#> 2 0.100(11) -0.00059  0.089  0.11
#> 
#>    correlation matrix of the fit parameters
#> 
#>           X1          X2         X3          X4
#> 1  1.0000000 -0.92547315 -0.6474289 -0.10607865
#> 2 -0.9254732  1.00000000  0.5140099 -0.08848184
#> 3 -0.6474289  0.51400991  1.0000000  0.18088613
#> 4 -0.1060786 -0.08848184  0.1808861  1.00000000
#> 
#>  estimates for x-values with errors, bootstrap bias and 68% confidence interval
#> 
#>           x     bias ci16 ci84
#> 1  1.00(10) -0.00086 0.90  1.1
#> 2 2.000(82)  -0.0016  1.9  2.1
#> 
#>    chi^2 and fit quality
#> chisqr / dof = 5.491311e-16 / 1 = 5.491311e-16 
#> p-value 1 
```

Perhaps it is easy to spot, but right now I do not see what is missing here. The examples need to be fixed to demonstrate that the code works, but I need to take a fresh look at this.

martin-ueding:
Trying to get rhdf5 to build on Travis CI turned out to be horrible and I have not succeeded. So I rolled back, and now it cannot find the data any more:

```
> data(plaq.sample)
Warning in data(plaq.sample) : data set ‘plaq.sample’ not found
> plaq.boot <- bootstrap.analysis(plaq.sample, pl=TRUE)
Error in bootstrap.analysis(plaq.sample, pl = TRUE) : 
  object 'plaq.sample' not found
Execution halted
```

This makes no sense, that has not been a problem before. The examples are run in descending order, so this had been running well before.

martin-ueding:
After rebasing and removing all these commits, it still fails at not being able to find `plaq.sample`, even though it is in `data/`, in `data/datalist` and even lazy loading is active. Do you have an idea?

martin-ueding:
I do not get it. The file is there, has not been touched in ten years. Also on Travis CI a call to `data()` lists exactly that data set. I have made a [Stack Overflow question](https://stackoverflow.com/questions/53283408/dataset-cannot-be-loaded-although-data-lists-it), hopefully somebody can shed some light on this.

martin-ueding:
Don't add `.Rdata` to the `.Rbuildignore`, otherwise these things happen.

martin-ueding:
I seem to have fixed the issues, now we have a failure because the fit in the example for `fit.cosh` fails to converge. @j-ostmeyer, could you take a look at this?

kostrzewa:
> Don't add .Rdata to the .Rbuildignore, otherwise these things happen.

Good to know!

j-ostmeyer:
From my point of view the branch is ready to merge.

PR #85, opened 2018-10-30
Omeas templates

kostrzewa:
Template files for driver scripts to run online measurements and produce RMarkdown reports.


martin-ueding:
It seems that mostly new functionality was added. Shall we just merge this?

PR #86, opened 2018-11-11
Jackknife as bootstrap

martin-ueding:
Most operations on the bootstrap samples are just an application to the original value and each bootstrap samples. The actual nature of the samples is not important when say solving the GEVP or evaluating the zeta function. This opens an easy path for implementing jackknife on almost all functions: We just put the jackknife samples into the `cf.tsboot$t` field of the `cf` objects. Most functions will not notice any difference.

Problematic are areas where errors are computed, say in `plot.cf`. Here we have invocations of `sd` scattered through the codebase. For bootstrap we need a different error computation function, which is just a scaled version of `sd` actually. The `cf_boot` mixin now contains the field `error_fn` which is just `sd` for bootstrap samples and our own function for jackknife samples. All functions where I could find `sd` now use `cf$error_fn` to compute errors.

This change needs to be done for all future edits as well, one must not use `sd` as the samples could very well be jackknife samples. In order to find `sd` it is very helpful to use the (Perl) regular expression `\bsd\b` or in Vim `\v<sd>`.

It would be nice to rename “boot” to “samples” everywhere, but that will likely break all analysis code. So I suggest we keep the name “boot” but then have to be conscious that it does really mean “samples” in a general way.

As Markus and I have used this trick in the analysis already, we believe that this approach works. There are two issues left:

1. Using `cov(bootstrap_samples)` does not do the right thing for jackknife samples, one needs to add the scaling factor just as with the regular standard deviation. There are a couple of calls to `cov` in the chi² fit, so we need to add `cf$cov_fn` as well. This is straightforward to do.

2. In `master`, jackknife and bootstrap samples are two different entities stored under different names. With this unification a straightforward “jackknife after bootstrap” procedure will not work, it would need to have another name. We might be able to just have “samples” and “samples of second order”, not further specifying what type they are. This way one could still do jackknife after bootstrap but has jackknife support in the core parts of the hadron library.

Markus and I already use this branch in our analysis, so I would favor to have this merged into `master` soon.


martin-ueding:
Now I have introduced the `cov_fn` field to the `cf_boot` class. This allows to use the appropriately normalized covariance also for jackknife samples. I have tried to use that in `matrixfit`, but that is not sufficient as the covariance matrix computed is ignored anyway:

```r
  CovMatrix <- NULL
  # we always use the boostrap samples to estimate the covariance matrix 
  CovMatrix <- cf$cov_fn(cf$cf.tsboot$t[,ii])
  
  ## for uncorrelated chi^2 use diagonal matrix with inverse sd^2
  M <- diag(1/CF$Err[ii]^2)
  if(useCov) {
    ## compute correlation matrix and compute the correctly normalised inverse
    ## see C. Michael hep-lat/9412087
    M <- try(invertCovMatrix(cf$cf.tsboot$t[,ii], boot.l=cf$boot.l, boot.samples=TRUE), silent=TRUE)
    if(inherits(M, "try-error")) {
      if( autoproceed ){
        M <- diag(1/CF$Err[ii]^2)
        warning("[matrixfit] inversion of variance covariance matrix failed, continuing with uncorrelated chi^2\n")
        useCov <- FALSE
      } else {
        stop("[matrixfit] inversion of variance covariance matrix failed!\n")
      }
    }
  }
```

You can see that `CovMatrix` is computed, but then it is just copied into the return value; no usage in between. The `invertCovMatrix` function uses `cov`, so one needs to pass that parameter through.

kostrzewa:
There is a good reason for this, at least there used to be. For noisy data, it can be shown that the covariance matrix can be approximated, which used to be done in invertCovMatrix. (see the reference noted in matrixfit). I think, however, that this has since been superceded.

martin-ueding:
I have found another invocation of `cov` and replaced it. Now everything should be in place.

Bartek, Markus and I discussed whether this approach is the one that we want to take. My alternative proposal was to make the samples some R6 class which then has member functions for everything that needs knowledge of the nature of the samples:

- Error computation
- Covariance computation
- Filling NA values

Then all functions need to be refactored to be a transformation on a single dataset, then we have a new `sample_apply` function which just applies the function to the original data (`cf0`) and then onto each of the samples. If there are more tricks that are done (covariance estimated from samples, filling of NA effective mass values), these need to be expressed with the methods of the sample class.

This would mean a change to all existing functions, but while we are at it we can make `cf$samples` a list of such classes and then have arbitrary many sample types in parallel.

I think that the hadron library does not need to support more than one type of samples at the same time, the paramval-framework can take care of this. We use this in the Lüscher-analysis such that the nature of the samples is just a parameter; we can do comparison plots just like with every other parameter.

This pull request seems to be no hindrance in this other approach. When considering bootstrap only, nothing has changed, really. For jackknife we go from not supported to supported with the caveat that one has to double-check that all `boot` variables are actually handled as general samples. So although this is not really the way that one would design it from a fresh start, I consider it an incremental improvement.

From my side we can merge this into `master`. @kostrzewa said that he wanted to carefully look at it. Does anyone else have objections?

R/cf.R
```diff
@@ -329,7 +360,7 @@ bootstrap.cf <- function(cf, boot.R=400, boot.l=2, seed=1234, sim="geom", endcor
   return(invisible(cf))
 }
 
-jackknife.cf <- function(cf, boot.l=2) {
+jackknife.cf <- function(cf, boot.l = 2) {
```
kostrzewa: can we default to "leave out 1" ?

R/effectivemass.R
```diff
@@ -174,7 +176,7 @@ fit.effectivemass <- function(cf, t1, t2, useCov=FALSE, replace.na=TRUE, boot.fi
   }
 
   # cf here is a bootstrapped effective mass and $t is the matrix of bootstrap samples
-  CovMatrix <- cov(cf$t[,ii])
+  CovMatrix <- cf$cf$cov_fn(cf$t[,ii])
```
kostrzewa: This is one of the reasons why I would prefer to have the effective mass as a mixin to the cf class. That way, the naming would be consistent and you would not need to refer to ```cf$cf``` here. I'm not saying that you should change this now, but it would make for a nicer interface in the long run in my opinion.

R/invertCovMatrix.R
```diff
@@ -1,4 +1,4 @@
-invertCovMatrix <- function(cf, boot.l=1, boot.samples=FALSE) {
+invertCovMatrix <- function(cf, boot.l=1, boot.samples=FALSE, cov_fn = cov) {
```
kostrzewa: very good, this continues to work as it used to

R/cf.R
martin-ueding: Of course, this was just the default before. It is changed now.

R/effectivemass.R
martin-ueding: I see it the other way. The effective mass *has* as a correlation function but it *is* not one. The big problem in this function is that the parameter which is of type `effmass` is called `cf`. If we called if `effmass`, then `effmass$cf$cov_fn` would be okay.

R/matrixfit.R
```diff
@@ -561,8 +561,8 @@ plot.matrixfit <- function(mfit, plot.errorband=FALSE, ylim, xlab="t/a", ylab="y
     if(mfit$model == "shifted") y <- pars[2]*pars[3]*( exp(-pars[1]*(tx-deltat/2)) - sgn*exp(-pars[1]*(T-(tx-deltat/2))))
     else if(mfit$model == "pc") y <- pcModel(par=par[1:3], t=tx, T=T, reference_time=mfit$reference_time)
     else y <- 0.5*pars[2]*pars[3]*( exp(-pars[1]*tx) + sgn*exp(-pars[1]*(T-tx)))
-    yp <- rep(1, times=length(tt))
     
+    # yp is the physical exponential in case we want to look at the ratio plot
```
kostrzewa: does this work? shouldn't yp exist in the scope outside of the if statement or does it already and I just don't see it here?

martin-ueding: I am not sure what exactly you mean. R has function scope, not lexical scope. So we can define variables within an `if`-statement, even though that would not work in C++.

Also this block is the first and only definition of `yp` in that function (in vim: call `:set hlsearch`, put the cursor onto it and press `*` or just search `/\v<yp>`). So I think that the `else` was indeed needed here.

kostrzewa: You're right of course, but why have I had this fail before?

kostrzewa: yp was defined further above and then overwritten in the original implementation, as is often done in hadron

kostrzewa: As long as it works I'm happy.

martin-ueding: > `yp` was defined further above and then overwritten in the original implementation, as is often done in hadron.

I see the point of this style of programming in C/C++, but when I have function scope I prefer to do it this way. As it is here I read it as “`yp` is this or that.”, the other style reads as “`yp` is that. Except in this case.”. In C++ I really like `auto const yp = … ? … : …;` just because I can put a `const` into it. I guess my Haskell excursion made me think of variables as constants and now I want to avoid changing values.

martin-ueding:
Can this be merged now?

PR #88, opened 2018-11-20
Refactor matrixfit models

martin-ueding:
@NikSchlage wants to add another model to use with `matrixfit`. I have the strong feeling that we need to refactor `matrixfit` to make the models independent of each other. Basically I see the following structure:

screenshot_20181120_173128

We have started by introducing an R6-class with the models and using inheritance for this. Since we already have `bootstrap.nlsfit`, we will use that. The `new_matrixfit` function will just take the data, instantiate a model and pass that along to `bootstrap.nlsfit`.

We already submit a pull request such that you can look at our work in an early stage.


kostrzewa:
Phi4model is a multi-exponential, no? Should probably be named as such. Are you sure matrixfit is the right place to make this generalisation? Generally we assume here just various amplitudes but the same mass and a functional form which differs at most in the sign of the time-reversed contribution. If I understand correctly, phi4model is not this kind of factorising fit, right?

martin-ueding:
The current name is “NParticleModel”, but @NikSchlage needs to shed some light on the exact form of the model.

> Generally we assume here just various amplitudes but the same mass and a functional form which differs at most in the sign of the time-reversed contribution. If I understand correctly, phi4model is not this kind of factorising fit, right?

If that is the bar, the `pc` model should not have been included in `matrixfit` in the first place because it can only do 1×1 fits. Therefore this already is a model which is not factorizing, yet it is included in there.

From what I have seen in Niko's attempt to add it to the current state of `matrixfit`, it can also be expressed as a factorizing fit. We planned to implement it as such as well.

> Are you sure matrixfit is the right place to make this generalisation?

Therefore I think it is the right place.

---

I have added a unit test framework (testthat) and wrote a first test with the shifted model. The other models will follow. We can then test the existing and new code against each other and hopefully I can then sell you the changes we are making.

martin-ueding:
I have completed the three models and added test cases. The tests fail due to numeric differences, otherwise it seems all good. This means that somehow the fit or the model is wrong but the structure itself is there.

martin-ueding:
Now all the three existing models give the same results. There were mostly sign errors and factors of 2 missing. We have to see that it builds on Travis CI, but then we are ready for a review.

R/new_matrixfit.R
```diff
@@ -0,0 +1,467 @@
...
+
+ShiftedModel <- R6Class(
+  'ShiftedModel',
+  inherit = MatrixModel,
+  public = list(
+    initialize = function (time_extent, parind, sign_vec, ov_sign_vec, delta_t) {
+      super$initialize(time_extent, parind, sign_vec, ov_sign_vec)
+      self$delta_t <- delta_t
+    },
+
```
martin-ueding: There should be no whitespace here for consistency.

R/new_matrixfit.R
```diff
@@ -0,0 +1,467 @@
...
+    },
+    reference_time = NA
+  )
+)
+
+
+## 3 particle correlation function in phi^4 theory:
+## C_3 (t) = 0.5*A_1^(3)*A_2^(3)*(exp(-E_3*(T-t))+exp(-E_3*t)) (3-particle correlator)
+##            + 0.5*A_3^(3)*A_4^(3)*exp(-(2/3)*E_3*(T/2)) * (exp(-(1/3)*E_3*(T-t)) + exp(-(1/3)*E_3*t)) (thermal pollution)
+
```
martin-ueding: This should become a Roxygen comment:

```r
#' 3 particle correlation function in phi^4 theory:
#'
#' @description
#' \deqn{C_3 (t) = 0.5*A_1^(3)*A_2^(3)*(exp(-E_3*(T-t))+exp(-E_3*t))}
…

File: R/new_matrixfit.R [hunk 3]
```diff
@@ -0,0 +1,467 @@
...
+  'Phi4Model',
+  inherit = MatrixModel,
+  public = list(
+    initialize = function (time_extent, parind, sign_vec, ov_sign_vec, delta_t) {
+      super$initialize(time_extent, parind, sign_vec, ov_sign_vec)
+      self$delta_t <- delta_t
+    },
+    
+    prediction = function (par, x) {
+
```
martin-ueding: Newline at the beginning of the function is unusual, other whitespace should be consistent.

R/new_matrixfit.R
```diff
@@ -0,0 +1,467 @@
...
+  } else if (model == 'n_particles') {
+    model_object <- Phi4Model$new(cf$Time, parind, sign.vec, ov.sign.vec)#cf$n_particles)
+  }
+  
+  if (missing(par.guess)) {
+    par.guess <- model_object$initial_guess(CF$Cor, parlist, t1, t2)
+  }
+
+  args <- list(fn = model_object$prediction,            # function
+               gr = model_object$prediction_jacobian,   # gradient
```
martin-ueding: I'd like to remove those comments, but they don't really hurt.

martin-ueding:
The `testthat` now run (and succeed) on Travis CI.

martin-ueding:
And as can be seen on Travis CI, the tests do not succeed yet:

```
── 1. Error: parind_1 (@test_parlist.R#21)  ────────────────────────────────────
argument "length_t" is missing, with no default
1: make_parind(parlist, length_time) at testthat/test_parlist.R:21
2: array(1, dim = c(n_corr * length_t, 2))
── 2. Error: parind_4 (@test_parlist.R#33)  ────────────────────────────────────
argument "length_t" is missing, with no default
1: make_parind(parlist, length_time) at testthat/test_parlist.R:33
2: array(1, dim = c(n_corr * length_t, 2))
```

This is because you have added a (not really needed) parameter to the function signature and now the tests fail.

kostrzewa:
Since this is being worked on fervently, would it make sense to also think about how to deal with correlation functions which cannot be symmetrised (when open or fixed boundary conditions in time are used, for example) or which has data only on a subset of all time slices?

R/new_matrixfit.R
```diff
@@ -489,8 +489,13 @@ make_parind <- function (parlist, corr_matrix_size, length_t, summands = 1) {
   n_corr <- corr_matrix_size
   ## parind is the index vector for the matrix elements
   parind <- array(1, dim = c(n_corr*length_t, 2))
+  parind_aux <- array(1, dim = c(n_corr*length_t, 2))
```
martin-ueding: This fixes the cases for a signal plus *one* pollution. What about having multiple pollutions? That is what the `summands` parameter was thought for.

R/new_matrixfit.R
```diff
@@ -0,0 +1,501 @@
...
+    reference_time = NA
+  )
+)
+
+#' 3 particle correlation function in phi^4 theory
+#'
+#' @description
+#' 3 particle correlator with thermal pollution term:
+#' \deqn{C_3 (t) = 0.5*A_1^(3)*A_2^(3)*(exp(-E_3*(T-t))+exp(-E_3*t)) + 0.5*A_3^(3)*A_4^(3)*exp(-(2/3)*E_3*(T/2)) * (exp(-(1/3)*E_3*(T-t)) + exp(-(1/3)*E_3*t))}
+Phi4Model <- R6Class(
```
martin-ueding: I still think that this model is not inherent to φ⁴ theory and should instead be named like `NParticleModel` or so.

R/new_matrixfit.R
```diff
@@ -0,0 +1,501 @@
...
+        j <- which(self$parind[,3]==i)
+        zp3 <- -self$ov_sign_vec[j]*par[self$parind[j,4]] * 
+               exp(-(par[1]*(2/3))*(self$time_extent/2)) * 0.5*(exp(-(par[1]*(1/3))*(x[j]-self$delta_t/2)) +
+               self$sign_vec[j]*exp(-(par[1]*(1/3)) * (self$time_extent-(x[j]-self$delta_t/2))))
+        
+        zp4 <- rep(0, length(zp))
+        j <- which(self$parind[,4]==i)
+             zp4 <- -self$ov_sign_vec[j]*par[self$parind[j,3]]*
+             exp(-(par[1]*(2/3))*(self$time_extent/2)) * 0.5*(exp(-(par[1]*(1/3))*(x[j]-self$delta_t/2)) +
+             self$sign_vec[j]*exp(-(par[1]*(1/3))*(self$time_extent-(x[j]-self$delta_t/2))))
```
martin-ueding: We have the almost same code here four times. Could that be simplified?

R/new_matrixfit.R
```diff
@@ -489,8 +489,13 @@ make_parind <- function (parlist, corr_matrix_size, length_t, summands = 1) {
   n_corr <- corr_matrix_size
   ## parind is the index vector for the matrix elements
   parind <- array(1, dim = c(n_corr*length_t, 2))
+  parind_aux <- array(1, dim = c(n_corr*length_t, 2))
   for (i in 1:n_corr) {
     parind[((i-1)*length_t+1):(i*length_t), ] <- t(array(parlist[, i] + 1, dim = c(2, length_t)))
+    parind_aux[((i-1)*length_t+1):(i*length_t), ] <- t(array(parlist[, i] + 1, dim = c(2, length_t)))
```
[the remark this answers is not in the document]
kostrzewa: because ```parind``` is grown iteratively below and the original ```parind``` is required. One could of course refer to the first couple of columns of ```parind``` in the loop below for the same effect.

new_matrixfit.Rmd
```diff
@@ -0,0 +1,71 @@
+```{r setup}
```
kostrzewa: shouldn't this be in the vignettes or something?

martin-ueding: In principle yes, but that is just the exploratory notebook to test the functions. I did not consider it a vignette that shows off the package, so I did not want to include it there yet.

NikSchlage:
Now the tests in test_parlist.R succeed. The problems were caused by the target matrix. To repair it, I changed the elements c(rep(c(2, 2, 2, 3, 3, 2, 3, 3), each = length_time)) in line 35 to c(rep(c(2, 2, 3, 3, 2, 3, 2, 3), each = length_time)) and now everything works fine.

DESCRIPTION
```diff
@@ -15,6 +15,7 @@ Maintainer: Carsten Urbach <urbach@hiskp.uni-bonn.de>
 SystemRequirements: Gnu Scientific Library version >= 1.8
 Description: Toolkit to extract hadronic quantities from Lattice QCD simulations. It contains functionality for IO, plotting, bootstrap and jackknife resampling, fitting, GEVP solving, error and autocorrelation estimation as well as other areas.
 Imports: boot, abind
+Depends: boot, R6
```
kostrzewa: I guess this should now just be in "Imports", right?

martin-ueding: I had to read [this](http://r-pkgs.had.co.nz/description.html) yet another time, and I am not certain that I got all the details right. But I think that having it in `Imports` is the right thing, there is no reason to use `Depends` for us.

The `R6` is somewhat special as we need that during compilation time. From [this footnote](https://adv-r.hadley.nz/r6.html#fn51) it seems that `Imports` is just fine.

NikSchlage:
The new NParticleModel (based on SingleModel) is implemented now. At the moment the fit does not converge (exceptions: 1 and 2 particle case without pollution terms), probably due to the fact that the priors are still missing.
I would be pleased if someone could have a look at those updates.

R/new_matrixfit.R
```diff
@@ -0,0 +1,460 @@
...
+                          fit.method = "optim",
+                          autoproceed = FALSE,
+                          par.guess,
+                          every,
+                          ...
+                          ) {
+  stopifnot(inherits(cf, 'cf_meta'))
+  stopifnot(inherits(cf, 'cf_boot'))
+  
+  if (model == 'pc') {
```
kostrzewa: I guess there should be a check for the number of observables in the cf too, right? Otherwise the logic for the 'pc' model below does not make sense.

NikSchlage: Thank you for your quick reply - I hope that the new checks are okay.

PR #90, opened 2018-11-26
Weight & shift with deltat != 1

maowerner:
I only included a parameter in `removeTemporal.cf` which is passed through to `takeTimeDiff.cf` 

Could someone review the weight exponential correction factors in lines 106-128? Additionally  does the weight.factor in line 147 need to be adjusted? I am not sure what this is.


maowerner:
I just realized, that my local changes to the rho analysis are also in the pull request. Should I take them out? They don't hurt anyway, do they?

R/removeTemporal.cf.R
```diff
@@ -10,19 +10,26 @@
...
-removeTemporal.cf <- function(cf, single.cf1, single.cf2,
-                              p1=c(0,0,0), p2=c(0,0,0), L,
-                              lat.disp=TRUE, weight.cosh=FALSE) {
+removeTemporal.cf <- function(cf, 
+                              single.cf1, 
+                              single.cf2,
+                              p1=c(0,0,0), 
+                              p2=c(0,0,0), 
+                              L,
+                              deltat = 1,
```
kostrzewa: please place new parameters at the end

kostrzewa:
As discussed, analytical review is required of the effective mass and any other routine which might use weight.factor. Further, I don't think that the effective mass (weighted or shifted) even works if the assumed functional dependence is not cosh.

martin-ueding:
I have derived how the weighted model should look like: [Weighted_Model.pdf](https://github.com/HISKP-LQCD/hadron/files/2831954/Weighted_Model.pdf)

I think that the current implementation of `removeTemporal.cf` is just wrong for the cosh-case because it assumes that `cosh(x)^{-1} = cosh(-x)`, which is only valid for `|x| \gg 1`. In the derivation Markus just divide by it.

Could you please have a look and tell me whether the thing that I assign to `model` is suitable for implementation in the correlator fit and the effective mass?

PR #93, opened 2018-12-13
Fix readbinarycf

martin-ueding:
I broke the readbinarycf and since there was no test, this was not caught. I have now fixed it as I needed to use it again.


kostrzewa:
automatic testing is great, but if travis keeps breaking stuff every couple of weeks, I have to say that it's quite frustrating...

urbach:
I agree!

PR #94, opened 2018-12-14
add support for averaging and striding to readcmifiles and use that t…

kostrzewa:
…o support multi-sample online measurements in analysis_online


kostrzewa:
Another band-aid on top of the more and more contrived analysis of online measurements. I have no idea when I fill find the time to really clean this up...

kostrzewa:
Since there are no tests for this at present and I've (manually) tested the changes on several sets of online measurements, I'll pull this in for now.

PR #98, opened 2018-12-20
Remove call to `data()` in example

martin-ueding:
In order to diagnose a build problem on Travis CI where the data could
not be found, we had added this as a diagnostic. It turned out that all
`.Rdata` files had been excluded in the `.Rbuildignore` due to a faulty
regular expression. This diagnostic is no longer needed and just
clutters the example.


martin-ueding:
Yes, I have just checked, this was the only place where I added it for debugging.

Issue #97, opened 2018-12-19
Add possibility of bootstrapped auxiliary parameters in bootstrap.nlsfit


martin-ueding:
The `bootstrap.nlsfit` function currently accepts bootstrap samples for the independent and dependent variable as well as a single closure which describes the model. This is become limiting.

For the Breit-Wigner fit in the ρ-project we have the phase shift as dependent and CM energy as independent variable. The function however also depends on the pion mass, which we currently have to put into the closure. This means that the pion mass uncertainty is neglected during this fit. As this uncertainty is small, the damage likely is small.

I see four ways to go forward:

1. Pass the ID `r` of the bootstrap sample (and `0` or `NA` for the original data) to the closure. In the concrete example it will then take `pion_mass_val` or `pion_mass_boot[r]`. This is most flexible.

2. Pass an additional matrix with `boot.R` or `1 + boot.R` number of rows to `bootstrap.nlsfit`. The function then slices this matrix and passes the appropriate row to the fit closure. The disadvantage is that the closure then has to unpack this row and additionally this has to be of homogeneous data type.

3. Somehow cram the additional parameters to be additional independent or dependent variables. Fill up the other one with `NA`. Then in the closure carefully unpack this.

4. Do not alter `bootstrap.nlsfit` at all. Instead let the closure also have a copy of the bootstrap samples of the independent variable. The closure then searches for the current set of independent variables in the matrix of bootstrap samples, extracts the row number and figures out the sample number in this way.

I prefer number 1, though.

martin-ueding:
There is one caveat with the first two options: The signature of the fit function need to account for this additional parameter. It is easy to fix for users as they can just add `, ...` to their function such that all additional parameters that they do not use are ignored. I think that is an acceptable burden.

PR #100, opened 2018-12-20
 Pass the bootstrap id to the fit functions

martin-ueding:
This is a solution for Issue #97.

In the implementation there is one potential stumbling block: The
ellipsis `...` is used in the signature of `bootstrap.nlsfit` to capture
arbitrary auxiliary fixed arguments that will be passed to the model
function. Before the `...` were passed directly from `bootstrap.nlsfit`
to the function, now it is passed to `nls.lm`/`optim`, then to `wrapper`
and then to the actual model function. So basically the scope of the
`...` have changed and include the `boot_r`.


kostrzewa:
I need to think about this for a moment since I haven't used the function yet.

martin-ueding:
Indeed, this was wrong. I changed my mind about passing `NA` and would like to pass `0` for the original data instead this way a user can just do `c(orig, boot)[boot_r + 1]` to access it from a list of elements. With `NA` this would be more code.

Issue #95, opened 2018-12-18
Publishing hadron on CRAN


martin-ueding:
Via email there was the idea to publish hadron on CRAN. I am not sure whether the optional dependency of `rhdf5` from the BioConductor repository makes this impossible. Once we have figured that out, we can further discuss.

---

Collection of tasks:

- [x] The license is states as `GPL 3 or later` which is not automatically recognized. We should either just write `GPL-3` and state that other versions are fine in some text, or live with this.
- [x] Using `library` within the code is apparently something that should not be done, we should instead use `::` or `requireNamespace`. The problem with using `library` or `require` in package code is that it alters the search path for the user of the package. [It is recommended](https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Suggested-packages) to instead use `requireNamespace('foo')` and then use `foo::bar`.
- [x] The dependency to `rhdf5` is not declared in the `DESCRIPTION`. I had tried to fix that but then Travis CI would fail. The dependency should be in there, though.
- [x] There is `boot:::ts.array` somewhere in the code, which is a no-go. We must not depend on implementation details that are not exported.
- [x] Fix warnings about probable errors.
- [x] Fix warnings about S3 naming inconsistencies.
- [x] Fix inconsistencies between function signatures and their hand-written documentation pages.
- [x] Document all parameters of already documented functions.
- [x] Document all functions
- [x] Document all data sets.
- [x] The compilation flags in `PKG_CFLAGS`, `-O3 -Wall`, are not portable. This need to be changed, perhaps R already passes sufficient optimization flags such that this is not needed?
- [x] The `hadron.so` exports some elements from the standard library, this likely is a mistake and needs fixing.
- [x] Use only ASCII characters in R code.

martin-ueding:
Then we also have this one here:

> In principle, packages must pass R CMD check without warnings or significant notes to be admitted to the main CRAN package area.

Currently there are *many* warnings that need to be fixed before we get to this point. Even if we end up not this submitting to CRAN, fixing all the warnings will improve the check experience and likely also the code itself.

martin-ueding:
Once we have CRAN for deployment, we can remove all the generated files from the git repository, right?

martin-ueding:
Regarding the license, [this is their list](https://www.r-project.org/Licenses/). It seems that we shall use `GPL-3` and indicate in some text that later versions are applicable.

martin-ueding:
There are a bunch of warnings like these:

```
[...]

PR #101, opened 2018-12-20
Towards CRAN

martin-ueding:
Addressing all the issues that come up with `R CMD check .`.


martin-ueding:
This pull request of course addresses Issue #95.

martin-ueding:
I am trying to properly expose the `rhdf5` dependency. Travis CI seems to happily *attempt* to install it, but fails with this:

```
H5constants.c: In function ‘_H5constants’:
H5constants.c:64:82: error: ‘H5F_LIBVER_V18’ undeclared (first use in this function)
   int const_H5F_LIBVER[4]       = {  H5F_LIBVER_EARLIEST,   H5F_LIBVER_LATEST,   H5F_LIBVER_V18 ,  H5F_LIBVER_V110  };
                                                                                  ^
```

I recently had the exact same issue on my laptop. I fixed this by removing `Rhdf5lib` and letting it install again. On Travis CI it seems to install `Rhdf5lib` in version 1.2.1. On my laptop I have 1.4.2 installed, perhaps this version is just too old. `rhdf5` 2.26.1 on Bioconductor also lists `Rhdf5lib(>= 1.3.2)` as dependency, so error should not occur. And indeed Travis CI tries to install `rhdf5_2.26.1`.

So we need to find out why `Rhdf5lib` is installed in such an old version.

martin-ueding:
Fixing the version for `rhdf5` did not work, but also depending explicitly on `Rhdf5lib` from Bioconductor has worked. I feel that this should not be necessary, but at least it is building now and we properly specify all dependencies. Next up is using `release` instead of `3.8` as I do not want to get stuck with a particular version.

martin-ueding:
As this works we can already merge it and then add onto it, i.e. continuous integration. Therefore I removed the “do not merge yet” label.

See the top post in Issue #95 to see what work is still left.

R/computeacf.R
```diff
@@ -34,8 +34,8 @@ computeacf <- function(tseries, W.max, Lambda=100) {
 }
 
 ## generic function to plot an object of class "myGamma"
-plot.hadronacf <- function (Gamma, col = "black", ...)
-{
+plot.hadronacf <- function (x, col = "black", ...) {
```
kostrzewa: is this a requirement?

kostrzewa: I mean copying the argument.

man/fit.cosh.Rd
```diff
@@ -38,13 +38,6 @@ plot.coshfit(x, col.fitline="black", plot.mass=TRUE, plot.corr=FALSE, ...)
   \item{useCov}{
     Use the correlated chisquare. This works only for not too noisy data.
   }
-  \item{replace.na}{
```
kostrzewa: is there are reason for the disappearance here?

R/computeacf.R
martin-ueding: The requirement is that the argument is called `x`. I could have renamed it all through the function, but this way the chances of breaking something are smaller. We can rename it to `x` in the whole function independently of that.

man/fit.cosh.Rd
martin-ueding: The function does not have that argument, so I have updated the documentation accordingly.

So it might be the question whether the function should have that argument. I am against that, we should have a separate function that fills the NA values in an effective mass.

R/computeacf.R
kostrzewa: I see, I was just wondering since you've made replacements like this in several places.

man/fit.cosh.Rd
kostrzewa: I agree.

R/cosh_nlsfit.R
```diff
@@ -161,7 +161,8 @@ fit.cosh <- function(effMass, cf, t1, t2, useCov=FALSE, m.init, par, n.cosh=2, a
   return(invisible(effMass))
 }
 
-plot.coshfit <- function(effMass, col.fitline="black", plot.mass=TRUE, plot.corr=FALSE, ...) {
+plot.coshfit <- function(x, col.fitline = "black", plot.mass = TRUE, plot.corr = FALSE, ...) {
+  effMass <- x
```
kostrzewa: ***must*** the first argument really be named ```x```?

martin-ueding: Yes, I think that it is because one wants to support named arguments. Say a user writes the generic `plot(x = samplecf)`, then this will fail with the current `plot.cf` because it expects a first argument called `cf`:

```
> plot(x = samplecf)
Fehler in any(inherits(cf, c("cf_orig", "cf_boot", "cf_jackknife"))) : 
  Argument "cf" fehlt (ohne Standardwert)
```

Having different argument names might be nice in the context of the particular function, but it breaks named arguments in the large picture.

kostrzewa: I see, it does make it ugly in the function, especially because 'x' is used elsewhere for the horizontal coordinate...

martin-ueding: With the `effMass <- x` we give that a proper name and free up `x` for use in the function, so perhaps the current state is not that bad after all for this function?

martin-ueding:
Could we also get this merged, please? It will remove the output on Travis CI and therefore makes it easier to find actual errors.

R/computeacf.R
urbach: its required for the package for some reason...

R/cosh_nlsfit.R
urbach: I support this renaming...!

R/bootstrap.nlsfit.R
```diff
@@ -462,10 +477,13 @@ bootstrap.nlsfit <- function(fn,
...
 #' @param object object returned by \code{bootstrap.nlsfit}
 #' @param digits number of significant digits to print in summary or print.
+#' @param print.correlation Logical. Whether to show the correlation between of
+#' the fit parameters.
+#' @param ... ignored
 #'
 #' @export
 #' @family NLS fit functions
-summary.bootstrapfit <- function(object, digits=2, print.correlation=TRUE) {
+summary.bootstrapfit <- function(object, digits = 2, print.correlation = TRUE, ...) {
```
urbach: Shouldn't `...` be the second argument?

R/bootstrap.nlsfit.R
```diff
@@ -511,20 +529,16 @@ summary.bootstrapfit <- function(object, digits=2, print.correlation=TRUE) {
 #'
 #' @param x object returned by \code{bootstrap.nlsfit}
 #' @param digits number of significant digits to print in summary or print.
+#' @param ... Additional parameters passed to the `summary.bootstrapfit` function.
 #'
 #' @family NLS fit functions
-print.bootstrapfit <- function(x, digits=2) {
+print.bootstrapfit <- function(x, digits = 2, ...) {
```
urbach: same question...

R/computeacf.R
urbach: likewise, `...` the second argument?

martin-ueding:
[It seems](https://stackoverflow.com/questions/54096383/where-shall-one-put-the-ellipsis-in-s3-function-overloads) that there is no general rule and it depends a bit on whether one wants to force these arguments to be used by name only. From what I saw in the documentation in the generic `print`, `summary` and `plot` the ellipsis is usually last in the methods. In say `tibble::as_tibble` the ellipsis is mostly as second argument.

I think that using them as second argument means that our special additional arguments need to be passed explicitly, which should make it a bit safer.

R/matrixfit.R
```diff
@@ -413,14 +410,7 @@ matrixfit <- function(cf, t1, t2,
...
     if(any(names(cf) == "weighted")) {
-      if(cf$weighted) {
-        fitfn <- matrixChisqr.weighted
-        dfitfn <- dmatrixChisqr.weighted
-      }
-      else {
-        model <- "single"
-        warning("model chosen to be weighted, but weighting information not available in cf. Falling back to model=single\n")
-      }
+      stop('Weighted model is not implemented.')
```
kostrzewa: what?

kostrzewa: oh, indeed, this somehow never made it into the codebase...

R/boot_ts_array.R
```diff
@@ -0,0 +1,35 @@
+#' Copy of boot:::ts.array
+#'
+#' The function `ts.array` of the `boot` package is not exported, yet we depend
+#' on it. It is bad to depend on private functions of a library, so we copy it
+#' here in the current version such that the `boot` package is free to change
+#' it.
+boot_ts_array <- function (n, n.sim, R, l, sim, endcorr)
```
kostrzewa: @martin-ueding 
Sorry you're going to quickly on this. There's too much going on at the same time to follow this in full detail. I at least need two or three weeks, not a few hours. Did you check that this works correctly? I think the RNG state needs to be reset for this to give the correct indices in the "geom" case.

There is an exported wrapper for this available: ```boot.array```. I don't think there's any need for using the unexported function.

kostrzewa: @urbach As a result, I'm not even sure whether your use was correct.

kostrzewa: I guess that is actually taken care of. However, what happens when the implementation of ```tsboot``` does change how "geom" is implemented? Then this will break silently...

urbach:
> 
> 
> 
> > @@ -0,0 +1,35 @@
> +#' Copy of boot:::ts.array
> +#'
> +#' The function `ts.array` of the `boot` package is not exported, yet we depend
> +#' on it. It is bad to depend on private functions of a library, so we copy it
> +#' here in the current version such that the `boot` package is free to change
> +#' it.
> +boot_ts_array <- function (n, n.sim, R, l, sim, endcorr) 
> 
> @martin-ueding 
> Sorry you're going to quickly on this. There's too much going on at the same time to follow this in full detail. I at least need two or three weeks, not a few hours. Did you check that this works correctly? I think the RNG state needs to be reset for this to give the correct indices in the "geom" case.
> 
> There is an exported wrapper for this available: ```boot.array```. I don't think there's any need for using the unexported function.

hmm, but ```boot.array``` does not work for the block indices
generated by tsboot, does it?

urbach:
> 
> 
> 
> > @@ -0,0 +1,35 @@
> +#' Copy of boot:::ts.array
> +#'
> +#' The function `ts.array` of the `boot` package is not exported, yet we depend
> +#' on it. It is bad to depend on private functions of a library, so we copy it
> +#' here in the current version such that the `boot` package is free to change
> +#' it.
> +boot_ts_array <- function (n, n.sim, R, l, sim, endcorr) 
> 
> I guess that is actually taken care of. However, what happens when the implementation of ```tsboot``` does change how "geom" is implemented? Then this will break silently...

yes, thats indeed the case. We can, of course, ask them to export the
function and take this as a workaround for the time being.

kostrzewa:
> hmm, but ```boot.array``` does not work for the block indices generated by tsboot, does it?

Looking at the current source code, there is logic for all of that:

```
boot.array <- function(boot.out, indices=FALSE) {
#
#  Return the frequency or index array for the bootstrap resamples
#  used in boot.out
#  This function recreates such arrays from the information in boot.out
#
    if (exists(".Random.seed", envir=.GlobalEnv, inherits = FALSE))
        temp <- get(".Random.seed", envir=.GlobalEnv, inherits = FALSE)
    else temp<- NULL
    assign(".Random.seed",  boot.out$seed, envir=.GlobalEnv)
    n <- NROW(boot.out$data)
    R <- boot.out$R
    sim <- boot.out$sim
    type <- find_type(boot.out)
    if (type == "tsboot") {
#  Recreate the array for an object created by tsboot, The default for
#  such objects is to return the index array unless index is specifically
#  passed as F
        if (missing(indices)) indices <- TRUE
        if (sim == "model")
            stop("index array not defined for model-based resampling")
        n.sim <- boot.out$n.sim
        i.a <- ts.array(n, n.sim, R, boot.out$l,
			sim, boot.out$endcorr)
        out <- matrix(NA,R,n.sim)
        for(r in seq_len(R)) {
            if (sim == "geom")
                ends <- cbind(i.a$starts[r,  ],
                              i.a$lengths[r,  ])
            else
                ends <- cbind(i.a$starts[r,], i.a$lengths)
            inds <- apply(ends, 1L, make.ends, n)
            if (is.list(inds))
                inds <- unlist(inds)[seq_len(n.sim)]
            out[r,] <- inds
        }
    }
    else if (type == "censboot") {
#  Recreate the array for an object created by censboot as long
#  as censboot was called with sim = "ordinary"
        if (sim == "ordinary") {
            strata <- tapply(seq_len(n), as.numeric(boot.out$strata))
            out <- cens.case(n,strata,R)
        }
        else	stop("boot.array not implemented for this object")
    }
    else {
#  Recreate the array for objects created by boot or tilt.boot
        if (sim == "parametric")
            stop("array cannot be found for parametric bootstrap")
        strata <- tapply(seq_len(n),as.numeric(boot.out$strata))
        if (find_type(boot.out) == "tilt.boot")
            weights <- boot.out$weights
        else {
            weights <- boot.out$call$weights
            if (!is.null(weights))
                weights <- boot.out$weights
        }
        out <- index.array(n, R, sim, strata, 0, boot.out$L, weights)
    }
    if (!indices) out <- freq.array(out)
    if (!is.null(temp)) assign(".Random.seed", temp, envir=.GlobalEnv)
    else rm(.Random.seed, pos=1)
    out
}
```

urbach:
> > hmm, but ```boot.array``` does not work for the block indices generated by tsboot, does it?
> 
> Looking at the current source code, there is logic for all of that:

well, the docu states differently... I'd need to understand, how
```boot.array``` works. Then this would be the better choice, of course...

kostrzewa:
> > hmm, but ```boot.array``` does not work for the block indices generated by tsboot, does it? Looking at the current source code, there is logic for all of that:
> well, the docu states differently... I'd need to understand, how ```boot.array``` works. Then this would be the better choice, of course...

If I understand correctly, the documentation refers to the special case of ```tsboot``` with ```model```-based resampling, in which case it cannot know how the model function resampled the data. The paragraph on this is very dense and hard to parse because several concepts are listed in short order.

urbach:
> > > hmm, but ```boot.array``` does not work for the block indices generated by tsboot, does it? Looking at the current source code, there is logic for all of that:
> > well, the docu states differently... I'd need to understand, how ```boot.array``` works. Then this would be the better choice, of course...
> 
> If I understand correctly, the documentation refers to the special case of ```tsboot``` with ```model```-based resampling, in which case it cannot know how the model function resampled the data. The paragraph on this is very dense and hard to parse because several concepts are listed in short order.

when trying naively, ts.array and boot.array give different results...

R/boot_ts_array.R
martin-ueding: I believe that I did not change the working of the code, I just copied `boot:::ts.array` to `boot_ts_array` and change the function calls. This way we are isolated against internal changes in `boot`, nothing more.

kostrzewa: No, if the internal implementation in ```boot``` changes, this function will give wrong results...

urbach:
> > > > hmm, but ```boot.array``` does not work for the block indices generated by tsboot, does it? Looking at the current source code, there is logic for all of that:
> > > well, the docu states differently... I'd need to understand, how ```boot.array``` works. Then this would be the better choice, of course...
> > 
> > If I understand correctly, the documentation refers to the special case of ```tsboot``` with ```model```-based resampling, in which case it cannot know how the model function resampled the data. The paragraph on this is very dense and hard to parse because several concepts are listed in short order.
> 
> when trying naively, ts.array and boot.array give different results...

```boot.array``` does not know about blocks, as far as I can see. It
returns the indices of the single elements. ```ts.array``` instead
gives the starting indices of each block and its length in number of
elements. I have to work through ```jab``` again in order to see what
I can do about it. But I currently don't see how to get away with
```boot.array```.

In any case, I'd not export ```boot_ts_array``` in hadron!

--
Carsten Urbach
e-mail: curbach@gmx.de
        urbach@hiskp.uni-bonn.de
Fon   : +49 (0)228 73 2379
skype : carsten.urbach

kostrzewa:
> when trying naively, ts.array and boot.array give different results...

I'm not surprised as ```boot.array``` applies a transformation to what is returned by ```ts.array``` and gives the individual measurement indices as opposed to starting positions and block lengths. This is done for both ```fixed``` and ```geom``` (but differently, of course).

urbach:
> > when trying naively, ts.array and boot.array give different results...
> 
> I'm not surprised as ```boot.array``` applies a transformation to what is returned by ```ts.array``` and gives the individual measurement indices as opposed to starting positions and block lengths. This is done for both ```fixed``` and ```geom``` (but differently, of course).

Of course, it is not a surprise, but difficult to get back to the
block information...

kostrzewa:
> Of course, it is not a surprise, but difficult to get back to the block information...

Yes, I see the point. The ```boot``` package also implements jab (although I don't think the function is exported), but it seems that it doesn't do block-based jackknife for that, which I guess is not quite correct, right?

kostrzewa:
> Well, we have neither a guarantee that boot:::ts.array remains present nor that it remains consistent with the other functions.

I guess I prefer the function disappearing such that it throws a hard error... Of course, if the implementation changes, also the fields in the object returned by ts.boot might change, thus rendering the copy here similarly non-functional.

urbach:
> > Well, we have neither a guarantee that boot:::ts.array remains present nor that it remains consistent with the other functions.
> 
> I guess I prefer the function disappearing such that it throws a hard error... Of course, if the implementation changes, also the fields in the object returned by ts.boot might change, thus rendering the copy here similarly non-functional. 

Problem is that if we go for CRAN, it will not work to import the
functions.

martin-ueding:
We can revert that commit of mine and just see what the people at CRAN say about this. We can also state that we thought about adding a convenience copy of the function to make it better.

PR #104, opened 2018-12-26
in shift.cf, use drop=FALSE to support shifting correlation function …

kostrzewa:
…with a single measurement


kostrzewa:
Also added loops to support cf object with multiple observables and smearing types

PR #106, opened 2018-12-26
perform apparently forgotten docs update

kostrzewa:


kostrzewa:
@martin-ueding dunno if this was an accident

PR #107, opened 2019-01-02
`raw_cf` container

kostrzewa:
The ```raw_cf``` container enables us to deal with complex and matrix-valued correlation functions.


kostrzewa:
This is still work in progress and I haven't yet added the arithmetic overloads or explicit test code, but blocking, uwerr analysis and plotting work.

Subsetting for arbitrary-dimension arrays is a bit clumsy in R: one is forced to construct index matrices via judicious use of ```expand.grid``` and ```do.call```. When the array dimensions are known, regular subsetting works of course. It's not a big deal but it makes working with the data a bit unpleasant. One also has to be on guard for inadvertent drops in the tensor dimension.

R/raw_cf.R
```diff
@@ -0,0 +1,550 @@
...
+#' @title Construct tensor index set for the internal degrees of freedom
+#' @param cf `raw_cf` container
+int_idx_matrix.raw_cf <- function(cf){
+  stopifnot(inherits(cf, 'raw_cf_meta'))
+  stopifnot(inherits(cf, 'raw_cf_data'))
+  args <- list()
+  for( d in cf$dim ){
+    args[[length(args)+1]] <- 1:d
+  }
+  as.matrix(do.call(exapnd.grid, args))
```
martin-ueding: There is a typo here in `expand.grid`.

R/raw_cf.R
```diff
@@ -0,0 +1,550 @@
...
+#' @title Add Gamma-method error analysis to existing `raw_cf` container
+#' 
+#' @param cf Correlation function container of class 'raw_cf'
+#' @family raw_cf constructors
+#' 
+#' @export
+raw_cf_uwerr <- function(cf){
+  stopifnot(inherits(cf, 'raw_cf_data'))
+  stopifnot(inherits(cf, 'raw_cf_meta'))
+  cf$uwerr <- uwerr.raw_cf(cf)
```
martin-ueding: I do not quite understand why this has to be part of the `raw_cf` class. To me it seems that this is something that is computed *from* a `raw_cf`. The `uwerr` object might have a reference to the original `raw_cf` if one wants to this “audit trail”. I fear that, just as with the `cf` class, it becomes very broad and large.

kostrzewa: It doesn't have to be, but if one wants to have it in there, one can also attach it and use it when necessary.

kostrzewa: You can make the same argument about bootstrap or jacknife samples...

R/raw_cf.R
kostrzewa: Cheers, fixed.

R/raw_cf.R
kostrzewa: The reason one might have for attaching it is how expensive this can become for large matrix-valued correlation funtions.

martin-ueding: I also do not like the way that bootstrap samples are handled in `cf`. This `cf_boot` is a step forward, but the `cf` class *is* too many different things, still. I just realized that what I actually have in mind would be that `cf_orig` is a separate class and a `cf_orig` *has* a `cf_orig` instance. Then we have `obj$meta$…` for the metadata and `obj$cf` for the original data. The `cf_boot` would then *have* a `cf_meta` instance and *have* a `tsboot` object.

Regarding expensive computations: I would think that if something is expensive to compute one should then keep track of that (separate) variable.

Perhaps we have different use cases for the library in mind. With the parameter-value-framework it is easy to have various values and I prefer small focused objects to work with. Also as the analysis is done with various script files and parameters one always knows what some function has been called with, for instance I do not need the fit object to remember the fit range as they are parameters and therefore always available.

On the other hand I have seen the hadron library used in an interactive session where more or less one `cf` object has been massaged to be various things through the functions.

martin-ueding: I currently have this structure in my head:

- All the `cf_` mixins become mutually exclusive classes.
- The `cf_orig` object has a member `meta` of class `cf_meta` and a member `cf` with the original data.
- In `bootstrap.cf` an instance of `cf_orig` is required. The return value is an instance of class `cf_boot` which has the member `meta` copied from input as one unit. There is no original data copied into `cf_boot`.
- This would also mean that a function like `invalidate.samples` becomes obsolete as there is no “going back”, one just uses the object with the original data.
- Stuff like the principal correlator just have a member of type `cf_boot` and the `gevp_reference_time` and `correlator_id` directly.
- Forwarding things like metadata becomes one instruction because it is just one structured element to copy.

So the thing with the current mixin approach that I am not happy with is that I believe that we actually have a composition hierarchy but do that in a flat namespace of a single object with multiple classes.

kostrzewa: > So the thing with the current mixin approach that I am not happy with is that I believe that we actually have a composition hierarchy but do that in a flat namespace of a single object with multiple classes.

Yes, I see the point. Moving away from the present setup, however, means that one will have to 

1) either construct a container externally to keep things together (when working with these things interactively), thus having to implement all the stuff that the ```cf``` class currently takes care of automatically (such as invalidating samples under transformations).
2) keep track of many objects, which is annoying interactively

I agree that inside a framework, working with small objects is often easier. However, exploratory analysis requires lots of meta-information and is ideally done interactively to make quick progress, at least that has been my experience so far. Once everything is packed up into some framework, it becomes somewhat difficult to follow what one is actually doing. Case in point is the difficulty that we faced extracting a particular parameter combination from a paramvalf container...

martin-ueding: I probably have the parameter-value-framework in mind and see the framework aspects of hadron as superfluous and would aim for a slim toolkit instead. But indeed, having some framework aspects makes it easier to use interactively.

I do not believe that one has to construct an additional container. The various members of the various mixins just need to be structured into a composition hierarchy. The `cf_orig` is then the container which contains the `cf_meta` and the actual data. The `cf_boot` is the container which contains the `cf_meta` and the bootstapped data; the original data would be lost after a `takeTimeDiff.cf` anyway.

All of these proposed changes are major refactorings of the whole library, so I guess it is mostly an academic exercise to discuss this. For the `raw_cf` class it should be still possible to change the design, although that will then be different to the current `cf`. Okay, I guess there is no obvious right answer to this …

kostrzewa: > I do not believe that one has to construct an additional container. The various members of the various mixins just need to be structured into a composition hierarchy. The cf_orig is then the container which contains the cf_meta and the actual data. The cf_boot is the container which contains the cf_meta and the bootstapped data; the original data would be lost after a takeTimeDiff.cf anyway.

I tend to agree with you more and more on this...

kostrzewa:
I add ```abind``` to imports because it seems a bit silly to use ```requireNamespace``` only in the (for now) two functions that use it. If one wishes to use ```raw_cf```, ```abind``` is a strict requirement.

kostrzewa:
Okay, if there are no opposing views I would like to merge this in now to then follow up with stuff which makes use of ```raw_cf```.

kostrzewa:
I will add down-converters to ```cf``` at a later stage.

kostrzewa:
Okay, now I'm definitely done fixing issues for now.

PR #110, opened 2019-01-08
Reading routine for nissa correlation functions

pittlerf:


kostrzewa:
Okay, I need to rework this...

src/read_nissa_correlation.cpp
```diff
@@ -0,0 +1,204 @@
...
+   int r1idx,r2idx;
+   int N_muindex1,N_muindex2;
+   char filename[100];
+   CharacterVector gamma_names(61);
+   NumericVector gammastructures(61);
+   int sizem1=nmasses1.size();
+   int sizem2=nmasses2.size();
+   int sizer1=r1.size();
+   int sizer2=r2.size();
+   if ((sizem1 != sizem2) || (sizem1 != sizer1) || (sizem1 != sizer2)){
```
kostrzewa: surely having ```sizem1 == sizer1``` is wrong, I don't think this check does what you intended, right?

pittlerf: I wrote it in a way, that you can separately read in every r values for any given m. So I assume an input like (m1,r1,m2,r2). In this way you can decide for each m the r value you are interested in separately, Of coarse in practice we are interested in only ++, or --.

```suggestion
   if ((sizem1 != sizem2) || (sizem1 != sizer1) || (sizem1 != sizer2)){
```

kostrzewa: that's not what I mean, you're comparing ```sizem1``` with ```sizer1``` and your suggested change is exactly what you have already...

in practice we can also be interested in the off-diagonal combinations for flavour-singlets

pittlerf: Sorry, I just by fall clicked on this suggested change, what I wanted to check here is just really that all four arrays (m1,m2,r1,r2) have the same length.  More precisely I have not done the direct product of the parameters in the code, but required it for the input parameters.

kostrzewa:
This is now in #113

PR #113, opened 2019-01-08
Nissa reader

kostrzewa:
Complete rework of GH #110 to work without external mapping of file contents and with arbitrary m1 / m2 / r1 / r2 / smearing combinations


kostrzewa:
> Is that really needed? I had the impression that it automatically picks up the stuff from `DESCRIPTION`. It at least does so with the `R6` package that we need for the `new_matrixfit`.

Not sure, was just following the pattern that was already estalished.

src/RcppExports.cpp
```diff
@@ -0,0 +1,42 @@
...
+    Rcpp::traits::input_parameter< CharacterVector >::type file_basenames_to_read(file_basenames_to_readSEXP);
+    Rcpp::traits::input_parameter< CharacterVector >::type smear_combs_to_read(smear_combs_to_readSEXP);
+    Rcpp::traits::input_parameter< const unsigned int >::type nts(ntsSEXP);
+    Rcpp::traits::input_parameter< DataFrame >::type combs_to_read(combs_to_readSEXP);
+    rcpp_result_gen = Rcpp::wrap(read_nissa_textcf_kernel(file_basenames_to_read, smear_combs_to_read, nts, combs_to_read));
+    return rcpp_result_gen;
+END_RCPP
+}
+
+RcppExport SEXP alphas(SEXP, SEXP, SEXP, SEXP, SEXP);
```
kostrzewa: @urbach @martin-ueding @pittlerf   Rcpp::compileAttributes() generates these automatically as it goes through the codebase to extract any Rcpp functions to be exported. Not sure what to do about the manual definitions of these. They could actually now be removed I guess.

kostrzewa:
@urbach @pittlerf 
Jeez do I hate text files for the output of multiple correlators... Anyway, this works now and is general in the sense that it can read any of the two-pt data produced by Nissa (at theta=0 and with a single stochastic source per correlator) even if the set of spin combinations or the number of masses / r combinations stored in the file were to change.

kostrzewa:
@pittlerf can you give this a try? Although I can say that it "works for me", that's not a particularly good test.

kostrzewa:
Okay, so the move of ```R/gsl_fit.R``` has been approved by @urbach. This is ready to go in.

pittlerf:
> can

HI Bartek, I tried to test it, but I got an error message like:
could not find function "read_nissa_textcf_kernel"

pittlerf:
> > can
> 
> HI Bartek, I tried to test it, but I got an error message like:
> could not find function "read_nissa_textcf_kernel"

OK, I solved this. I had to do a Rcpp::compileAttributes("/qbigwork2/pittler/lqcdhadron/hadron/") and after that compiling RcppExports.cpp.

martin-ueding:
I found that just doing `devtools::load_all()` takes care of all these things.

kostrzewa:
> I found that just doing `devtools::load_all()` takes care of all these things.

Sure, but it should work nonetheless. All the required exports should be available and there should't be anything that one needs to do.

@pittlerf What did you do to try to install the package?

pittlerf:
> > I found that just doing `devtools::load_all()` takes care of all these things.
> 
> Sure, but it should work nonetheless. All the required exports should be available and there should't be anything that one needs to do.
> 
> @pittlerf What did you do to try to install the package?

 R CMD INSTALL -l /hadron/pittler/R/x86_64-pc-linux-gnu-library/3.4 /qbigwork2/pittler/lqcdhadron/hadron

pittlerf:
> > > I found that just doing `devtools::load_all()` takes care of all these things.
> > 
> > 
> > Sure, but it should work nonetheless. All the required exports should be available and there should't be anything that one needs to do.
> > @pittlerf What did you do to try to install the package?
> 
> R CMD INSTALL -l /hadron/pittler/R/x86_64-pc-linux-gnu-library/3.4 /qbigwork2/pittler/lqcdhadron/hadron

For just one entry of correlation function I got the following error:
new <- readnissatextcf(sprintf("/hiskp4/pittler/pleugamma/data/beta1.726/cA211a.53.24/%04d/mes_contr_2pts",1),"ll",48, temp, 48, 1,TRUE)

pittlerf:
> > > > I found that just doing `devtools::load_all()` takes care of all these things.
> > > 
> > > 
> > > Sure, but it should work nonetheless. All the required exports should be available and there should't be anything that one needs to do.
> > > @pittlerf What did you do to try to install the package?
> > 
> > 
> > R CMD INSTALL -l /hadron/pittler/R/x86_64-pc-linux-gnu-library/3.4 /qbigwork2/pittler/lqcdhadron/hadron
> 
> For just one entry of correlation function I got the following error:
> new <- readnissatextcf(sprintf("/hiskp4/pittler/pleugamma/data/beta1.726/cA211a.53.24/%04d/mes_contr_2pts",1),"ll",48, temp, 48, 1,TRUE)

Error in cf$cf[, (istart + 1):(ihalf - 1)] : 
  incorrect number of dimensions

kostrzewa:
> Sure, but it should work nonetheless. All the required exports should be available and there should't be anything that one needs to do.

I had forgotten to add ```R/RcppExports.R``` to the repo.

kostrzewa:
@martin-ueding I've added the running of ```Rcpp::compileAttributes()``` to the pre-commit hook. Do you agree with this?

martin-ueding:
> I've added the running of `Rcpp::compileAttributes()` to the pre-commit hook. Do you agree with this?

As our goal is to make installation with `R CMD install` or `install.packages('.', repo = NULL)` possible with a source checkout and without additional dependencies, I guess that we have to do that.

kostrzewa:
> As our goal is to make installation with R CMD install or install.packages('.', repo = NULL) possible with a source checkout and without additional dependencies, I guess that we have to do that.

Okay. I've also added calls to ```git status``` and ```git status -u``` to the output in case a difference was found.

urbach:
is @pittlerf approving this!?

pittlerf:
> is @pittlerf approving this!?

Yes, I have tested this on the plng data.

PR #120, opened 2019-01-09
Commit hook (second batch)

martin-ueding:
I just realized that `git diff` does not tell us whether there are any new files. So I now also make sure that the output of `git status -s` has not changed.

Additionally only symlinks that do not exist yet are created.

### DIFF
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -26,10 +26,10 @@ cleanup() {
 trap cleanup EXIT
 
 git diff > "$diff_old"
-git status --porcelain > "$status_old"
+git status --porcelain -u > "$status_old"
 Rscript -e 'devtools::document();'
 git diff > "$diff_new"
-git status --porcelain > "$status_new"
+git status --porcelain  -u> "$status_new"
 
 if ! diff -q "$diff_old" "$diff_new" &> /dev/null; then
   echo "${red}Generating the documentation has changed the tree, add them to the index and try to commit again.${reset}" >&2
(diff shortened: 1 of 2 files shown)


kostrzewa:
Shouldn't the status call be ```-us``` ?

kostrzewa:
I mean ```-u```

kostrzewa:
I have

```
git config --global status.showUntrackedFiles 
no
```

set, so ```git status -s``` does not show untracked files.

martin-ueding:
It seems that the default value for `-u` is `all`. But yes, with the change in the configuration we need to give that.

j-ostmeyer:
But -u shows a lot of things that we are not interested in.

hooks/pre-commit
```diff
@@ -13,19 +13,30 @@ reset="$(tput sgr0)"
...
   rm -f "$diff_old"
   rm -f "$diff_new"
+  rm -f "$status_old"
+  rm -f "$status_new"
 }
 
 trap cleanup EXIT
 
 git diff > "$diff_old"
+git status --porcelain > "$status_old"
```
kostrzewa: nope, doesn't work for me

kostrzewa: only if I add ```-u``` in addition does this work

martin-ueding:
Huh, that seems to be against the manual:

```
--porcelain[=<version>]
           Give the output in an easy-to-parse format for scripts. This is similar to the short output, but will
           remain stable across Git versions and regardless of user configuration. See below for details.

           The version parameter is used to specify the format version. This is optional and defaults to the
           original version v1 format.
```

Okay, then we also specify `-u`.

PR #122, opened 2019-01-10
Add seed argument to parametric bootstrap

martin-ueding:
The parametric bootstrap functions did not fix the seed, this meant that we did not get correlation when wanted. Also it made it not reproducible. I have introduced a `seed` argument to these functions. The swapping and restoring of the seed is taken from `bootstrap.cf` and moved into separate functions for reuse.

Supplying a default seed to these functions would mean that the result from parametric bootstrap of independent variables would become exactly correlated. The user would have to remember to use a different seed for each invocation. As this is subtle I have chosen to not fix the default seed but instead use the global seed. This means that by default the parametric bootstrap is not reproducible and once the user specifies the seed they have to understand the consequences.


kostrzewa:
Looks good!

Issue #123, opened 2019-01-10
tex.catwitherror fails for very small (< 1e-20) numbers


martin-ueding:
As discussed in the wake of PR #40, `tex.catwitherror` fails when using very small numbers as value. The error that one gets is this:

```
Error in prettyNum(.Internal(format(x, trim, digits, nsmall, width, 3L,  : 
  invalid 'nsmall' argument
```

It arises from the use of `%f` formatting (fixed point) for a number which is too small to be displayed within the number of printed digits:

```
nsmall: the minimum number of digits to the right of the decimal
          point in formatting real/complex numbers in non-scientific
          formats.  Allowed values are ‘0 <= nsmall <= 20’.
```

We will need to switch to exponential notation for these cases.

PR #124, opened 2019-01-10
Split off exponent for very large or small numbers

martin-ueding:
This fixes Issue #123.

I have added logic to split off the exponent for very large or small numbers. The following test cases probably best describe the behavior:

```r
    expect_equal(tex.catwitherror(1.23e-20, 0.45e-20, digits = 2, with.dollar = FALSE), '1.23(45)e-20')
    expect_equal(tex.catwitherror(1.23e-40, 0.45e-40, digits = 2, with.dollar = TRUE), '$1.23(45)\\cdot 10^{-40}$')

    expect_equal(tex.catwitherror(1.23e-40, digits = 2, with.dollar = FALSE), '1.2e-40')
```

I hope that this works for formatting arbitrary χ² values.

### DIFF
--- a/R/tex-utils.R
+++ b/R/tex-utils.R
@@ -29,6 +29,15 @@ tex.catwitherror <- function(x, dx, digits=1, with.dollar=TRUE, human.readable=T
   N <- 0
   threshold <- 10^(digits-1)
 
+  base <- round(log10(x[1]))
+  split_base <- abs(base) > 19
+  if (split_base) {
+    x <- x / 10**base
+    if (!missing(dx)) {
+      dx <- dx / 10**base
+    }
+  }
+
   if(missing(dx) && lx < 2) {
     if( is.na(x) ) x <- 0.0
     ## just a number without error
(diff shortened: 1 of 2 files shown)


martin-ueding:
It failed because the combination `tex.catwitherror(NA, 1.0)` used to give just `"0(1)"`. I have added a check for this.

But I do not think that this is the desired result. I would rather have `"NA(1.0)"` or just `"NA"` or maybe even an actual `NA`. What do you want to have in that case?

kostrzewa:
> good question, I have to think about it. But an error on NA does not make any sense...

I guess it would indicate that the computation of the central value had some issue while that of the error did not.

martin-ueding:
Yes, it does not make sense, but one of the `bootstrap.nlsfit` examples failed with

    Error in if (split_base) { : missing value where TRUE/FALSE needed

And the only way I could reproduce that was with an NA value and finite error:

```
> tex.catwitherror(NA)
Fehler in tex.catwitherror(NA) : 
  x must be a numeric vector with length > 0
Called from: tex.catwitherror(NA)
```
```
> tex.catwitherror(NA, NA)
Fehler in tex.catwitherror(NA, NA) : 
  x must be a numeric vector with length > 0
Called from: tex.catwitherror(NA, NA)
```
```
> tex.catwitherror(c(1.0, NA))
[1] "$1(0)$"
```
```
> tex.catwitherror(c(NA, 1.0))
Fehler in if (split_base) { : Fehlender Wert, wo TRUE/FALSE nötig ist
```

urbach:
> > good question, I have to think about it. But an error on NA does not make any sense...
> 
> I guess it would indicate that the computation of the central value had some issue while that of the error did not.

okay, agree...

martin-ueding:
It was something else: When taking the log of a negative real number one gets NaN, which gives `NA` in comparisons.

j-ostmeyer:
Do we want to round for splitting of the exponent or should we rather take the floor?

PR #132, opened 2019-01-14
reader for Cyprus CalcLoops format

kostrzewa:
This is not quite complete yet, will be done over the next few days.


kostrzewa:
I'm happy with this now. Seems to work for practical purposes. The spin projections have to be checked via @pittlerf's implementation, but that should not prevent this from being pulled in.

kostrzewa:
@gbergner @uwenger @agasbarro @marcuspetschlies  @Blons @FernandaSteffens To keep you folks up to date, we can now read and work with the Cyprus loop files. Btw: The transformation of accumulated loops to loops from individual sources seems to work fine, at least on the 24c48 lattice. A very quick first analysis of some pertinent quantities will follow soon.

R/cyprus_readutils.R
```diff
@@ -0,0 +1,253 @@
...
+#'              type loops from the Cyprus CalcLoops application.
+#' @param istoch Integer, index of the stochastic sample that the key should
+#'               be generated for.
+#' @param loop_type String, name of loop type. Allowed values:
+#'                  'Scalar', 'dOp'
+#' @param cid Integer, configuration number, internally produced by the CalcLoops
+#'            tool via the "trajectory" input flag. The default is '4' as this is
+#'            often not used in practice.
+#' @export
+cyprus_make_key_scalar <- function(istoch, loop_type, cid = 4){
```
urbach: does this function and the following ones need export?

R/cyprus_readutils.R
```diff
@@ -0,0 +1,253 @@
...
+cyprus_read_loops <- function(selections, files, Time, nstoch, accumulated = TRUE, legacy_traj = TRUE, verbose = FALSE){
+  rhdf5_avail <- requireNamespace("rhdf5")
+  dplyr_avail <- requireNamespace("dplyr")
+  if( !rhdf5_avail | !dplyr_avail ){
+    stop("The 'dplyr' and 'rhdf5' packages are required to use this function!\n")
+  }
+  if( verbose ){
+    tictoc_avail <- requireNamespace("tictoc")
+    if( !tictoc_avail ){
+      stop("Time reporting requires the 'tictoc' package!")
```
urbach: so, this is a hard dependence on this package, isn't it?

R/looptools.R
```diff
@@ -0,0 +1,106 @@
+#' @title compute two-point correlation function between quark loops
+#' @param loop_snk 'raw_cf' container with spin-projected quark loop
+#'                 at sink (see \link{loop_spin_project})
+#' @param loop_src 'raw_cf' container with spin-projected quark loop 
+#'                 at source (see \link{loop_spin_project})
+#' @return 'raw_cf' container with two-point function of these two quark loops.
+#'         In the calculation, both averaging over all source locations and
+#'         the average over all stochastic sample combinations are performed.
+#' @export
+loop_2pt <- function(loop_snk, loop_src){
```
urbach: does this replace `computeDisc`?

R/cyprus_readutils.R
martin-ueding: I think that this is similar to the HDF5 reading that we have. Most of the package can be used without these packages installed, so I think that this usage does not imply an `Imports` dependency but a mere `Suggests`.

urbach:
> +      stop("Time reporting requires the 'tictoc' package!")
> 
> I think that this is similar to the HDF5 reading that we have. Most of the package can be used without these packages installed, so I think that this usage does not imply an `Imports` dependency but a mere `Suggests`.

Suggest is fine. I was more after the fact that time measurement is
not really essential, while a HDF5 library is... Well, all on CRAN, so
should be fine after all, but...

R/cyprus_readutils.R
kostrzewa: No, only if one wishes to use verbose mode.

urbach:
> The problem is that the example files that I would have been able to
> produce up until recently were at least several tens of GB in
> size. Now that we can filter the .h5 files to have just the zero
> momentum data for the Scalar, say, adding example data to the
> https://github.com/HISKP-LQCD/hadron_example_data repository will be
> no problem.

perfect!

R/looptools.R
kostrzewa: Not necessarily. I just needed a function to handle the 'raw_cf' format to work comfortably with the data that I have since complex multiplications are not overloaded for 'cf'. 'raw_cf' kind of formalises how to deal with data that can generally be complex and has more than measurement and time dimensions. For the loops, for instance, one can have a "samples" dimension as well as two Dirac dimensions.

urbach:
> 
> 
> 
> > +#'                    the loop files were generated, set this to \code{TRUE} and the paths
> +#'                    will be constructed with 'conf_0004' as their root group.
> +#' @param verbose Boolean, output I/O time per file. Requires 'tictoc' package.
> +#' @return Named list of the same length as \code{selections} containg the loop data
> +#'         in the \link{raw_cf} format.
> +#' @export
> +cyprus_read_loops <- function(selections, files, Time, nstoch, accumulated = TRUE, legacy_traj = TRUE, verbose = FALSE){
> +  rhdf5_avail <- requireNamespace("rhdf5")
> +  dplyr_avail <- requireNamespace("dplyr")
> +  if( !rhdf5_avail | !dplyr_avail ){
> +    stop("The 'dplyr' and 'rhdf5' packages are required to use this function!\n")
> +  }
> +  if( verbose ){
> +    tictoc_avail <- requireNamespace("tictoc")
> +    if( !tictoc_avail ){
> +      stop("Time reporting requires the 'tictoc' package!")
> 
> No, only if one wishes to use verbose mode.

Ahh, you are right of course. Then this is fine!

R/looptools.R
kostrzewa: Since the input arguments for ```computeDisc``` are not actually real ```cf``` objects, it's also not quite straightforward to drop down from ```raw_cf``` to ```cf``` and then use ```computeDisc```. However, I think in practice one would load data into ```raw_cf``` for first analysis to see if everything looks kosher and then construct the necessary input for ```computeDisc``` from that. I think a renaming of ```computDisc``` might be good to clarify that his actually computes a two-point correlation.

R/cyprus_readutils.R
kostrzewa: I guess not, one could leave it out.

kostrzewa: Done.

PR #133, opened 2019-01-15
make raw_cf generics comply with required generics argument names

kostrzewa:


kostrzewa:
@martin-ueding has taken care of most of the rest already in the CRAN pull-requests

Issue #140, opened 2019-01-24
jackknife_cov(…, na.rm = TRUE) does not work


martin-ueding:
`jackknife_cov(…, na.rm = TRUE)` does not work as expected when there is a row of `NA` measurements. We would expect it to just discard that row of measurements and adjust the widths of the distributions accordingly. This vignette shows the problem:

[jackknife_cov_and_missing_values.pdf](https://github.com/HISKP-LQCD/hadron/files/2792166/jackknife_cov_and_missing_values.pdf)

PR #142, opened 2019-01-24
Fix na rm jackknife cov

matfischer:
Fixed bug at jackknife_cov. Issue #140


R/cf.R
```diff
@@ -146,16 +146,20 @@ jackknife_error <- function (samples, na.rm = FALSE) {
...
     factor <- 1.0
     
     if (is.null(y)) {
         N <- nrow(x)
         if (na.rm) {
-            na_values <- apply(x, 2, function (row) any(is.na(row)))
+            na_values <- apply(x, 1, function (row) any(is.na(row)))
             m <- sum(na_values)
-            x <- x[!na_values, ]
+            x <- x[!na_values,]
```
martin-ueding: That whitespace was intentional there.

kostrzewa: agreed, makes it clearer

R/cf.R
```diff
@@ -146,16 +146,20 @@ jackknife_error <- function (samples, na.rm = FALSE) {
...
+#'
 #' @export
 jackknife_cov <- function (x, y = NULL, na.rm = FALSE, ...) {
     factor <- 1.0
     
     if (is.null(y)) {
         N <- nrow(x)
         if (na.rm) {
-            na_values <- apply(x, 2, function (row) any(is.na(row)))
+            na_values <- apply(x, 1, function (row) any(is.na(row)))
```
martin-ueding: We want to throw out the rows that contain `NA` values. Therefore margin 1 (rows) is the one to keep.

martin-ueding:
I would have hoped that `x[!na_values, ]` would have failed if `nrow(x)` did not match `length(na_value)`. But R does not even issue a warning in this case:

```r
> (1:10)[1:3 < 3]
[1]  1  2  4  5  7  8 10
```

If the index vector is too short, it is just filled up with `TRUE`.

With NumPy this is a hard error:

```python
>>> import numpy as np
>>> np.arange(1, 11)[np.arange(1, 4) < 4]
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
IndexError: boolean index did not match indexed array along dimension 0; dimension is 10 but corresponding boolean dimension is 3
```

R/cf.R
```diff
@@ -146,14 +146,17 @@ jackknife_error <- function (samples, na.rm = FALSE) {
...
```
kostrzewa: Does this actually happen in practice using Jackknife samples? Leaving out a single measurement should, I would guess, not really lead to things suddenly becoming badly defined. It's a different matter for the bootstrap, I would say.

kostrzewa:
> I would have hoped that `x[!na_values, ]` would have failed if `nrow(x)` did not match `length(na_value)`. But R does not even issue a warning in this case:
> 
> ```r
> > (1:10)[1:3 < 3]
> [1]  1  2  4  5  7  8 10
> ```
> 
> If the index vector is too short, it is just filled up with `TRUE`.
> 
> With NumPy this is a hard error:
> 
> ```python
> >>> import numpy as np
> >>> np.arange(1, 11)[np.arange(1, 4) < 4]
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
> IndexError: boolean index did not match indexed array along dimension 0; dimension is 10 but corresponding boolean dimension is 3
> ```

Yes, the best thing to do is to use ```which(1:3 < 3)``` instead.

R/cf.R
martin-ueding: I am not sure what you are aiming at. Matthias has the problem that there are some `NA` values in the effective masses that Markus gave him.

For the bootstrap leaving out samples is no problem, the distribution is not affected by it, how well it is resolved. For jackknife the number of samples is intrinsically linked to the width of the distribution, therefore omitting samples makes one underestimate the error.

martin-ueding:
> Yes, the best thing to do is to use `which(1:3 < 3)` instead.

No, that would only make things worse if R would actually signal an error condition. I would want R to raise an error when one passes a logical index vector which does not match in size exactly. By using `which` you convert the logical index vector into an integer index vector. The length of that is arbitrary, it can be shorter or longer than the vector. If one was to do that in NumPy, one would lose the length check.

In R one can even pass a logical index vector which is longer than the vector. For every `TRUE` there one just gets `NA`. I dislike that R has defined output for strange inputs, I'd favor exceptions being raised.

R/cf.R
kostrzewa: > I am not sure what you are aiming at. Matthias has the problem that there are some NA values in the effective masses that Markus gave him.

But why would this even happen in a jackknife sample? My point is that I find it very unlikely that an effective mass works correctly on 251 measurements but then fails to compute correctly on a subset of 250 measurements. Are you sure that these effective masses use jackknife samples?

> For the bootstrap leaving out samples is no problem, the distribution is not affected by it, how well it is resolved. For jackknife the number of samples is intrinsically linked to the width of the distribution, therefore omitting samples makes one underestimate the error.

I know, that's not the origin of my question.

kostrzewa:
> > Yes, the best thing to do is to use `which(1:3 < 3)` instead.
> 
> No, that would only make things worse if R would actually signal an error condition. I would want R to raise an error when one passes a logical index vector which does not match in size exactly. By using `which` you convert the logical index vector into an integer index vector. The length of that is arbitrary, it can be shorter or longer than the vector. If one was to do that in NumPy, one would lose the length check.

I find that the ```which``` form does what you intended. You pass an integer index vector containing only the elements that you want.

> In R one can even pass a logical index vector which is longer than the vector. For every `TRUE` there one just gets `NA`. I dislike that R has defined output for strange inputs, I'd favor exceptions being raised.

I agree that the behaviour of logical index vectors is braindead.

kostrzewa:
> ex vector into an integer index vector. The length of that is arbitrary, it can be shorter or longer than the vector. If one was to do that in NumPy, one would lose the length check.

When an integer index vector is longer than the vector itself but contains only indices which exists, the result is a vector which is larger than the original vector. This is exactly what one wants to happen as it allows for very simple syntax if one is interested in repeating indices certain indices several times.

R/cf.R
matfischer: Maybe it's not the best place to discuss it here but principally you are right Bartek. The problem for me is, that I have to cut out time slices which are NA in my routine. There are some correlators/effectivemasses which have NA all over one column . If I cut out all rows which have NA then I have no covariance matrix anymore. That's why I ran my routine over all rho data cutting out all time slices which contained any NA in a column. Doing this I need to multiply the respective covariance matrix just by the global factor which is inserted at jackknife_cov(...).

The main problem for my routine is that it is dependent on the method which is applied to the correlators to obtain the effectivemass data. That's why we need to think of a method to handle the used/needed "x"-values (without NAs). I need df <- data.frame(x = x, y = y, dy = dy) and the inverse of the cov matrix which I handle separately to the function call which runs the whole routine. That would be the only "interface" between LuescherAnalysis and BayesianFitting.

I will try to finish the R package over the weekend such that we all can have a look at it and test it if you are interested.

kostrzewa: NA's in one or multiple **columns** are not a surprise. Hadron does not currently employ periodic boundary conditions in the computation of the correlator time derivative or the effective mass, such that one or multiple columns become undefined in the procedure. These columns should be **completely ignored** in any case and you should know **at the calling point of your routine**, that these columns are not to part of the dataset. In particular, removing **columns** from the bootstrap or jackknife matrix **must not change the jackknife error scaling factor**.

The situation we are discussing with @martin-ueding is a different one: when bootstrapping, the resampled correlator can behave so weirdly that the computation of the effective mass breaks down. Thus, on some bootstrap samples (**rows** of the bootstrap matrix, not columns), the effective mass might be NA for one or a few time slices. In the bootstrap case, the data is then simply resampled unless there are too many of these. In the jackknife case, resampling is not possible, so the treatment of the error and covariance must be adjusted.

I am very surprised that one would find a similar situation on jackknife samples, unless there are only very few measurements in total.

kostrzewa: > The main problem for my routine is that it is dependent on the method which is applied to the correlators to obtain the effectivemass data. That's why we need to think of a method to handle the used/needed "x"-values (without NAs). I need df <- data.frame(x = x, y = y, dy = dy) and the inverse of the cov matrix which I handle separately to the function call which runs the whole routine. That would be the only "interface" between LuescherAnalysis and BayesianFitting.

Yes that's a good interface.  Thus, at the calling point in your analysis (where you know which data you are passing to BayesianFitting), you should restrict yourself to the x values for which the effective mass of that particular type is well defined. This should certainly not be done inside of ```jackknife_cov```.

matfischer: Yes, you are absolutely right. I just wanted to say that cutting rows is not that important to "my" routine since I need to cut columns first. I just spotted the bug when I looked at `jackknife_cov' because I considered using it. I thought that one could extend this to cut out columns which are completely NA within `jackknife_cov`. But now it is clear to me why we do not want this at this function. However, I should use `jackknife_cov` after removing the NA cols. Thanks.

PR #147, opened 2019-01-28
Avoid warnings in plotwitherror

### LINKED ISSUE #144: plotwitherror gives warnings when the error is small
We thought that we fixed the zero-length arrow warnings in Issue #60, but this also occurs when the errors are very small. The two should both give a sensible plot:

```r
plotwitherror(0, 1, dy = 0.1, ylim = c(0, 2))
plotwitherror(0, 1, dy = 1e-10, ylim = c(0, 2))
```

The second one however gives the following warnings:

```
Warnmeldungen:
1: In arrows(x[rw], start, x[rw], end, length = arwhd.len, angle = 90,  :
  zero-length arrow is of indeterminate angle and so skipped
2: In arrows(x[rw], start, x[rw], end, length = arwhd.len, angle = 90,  :
  zero-length arrow is of indeterminate angle and so skipped
```

Apparently the length of the error needs to be sizable compared to the plot region for this to give no warning. I do not know how we could possibly know that in advance, expecially with `rep = TRUE`.

j-ostmeyer:
Resolves #144 .

### DIFF
--- a/R/plotutils.R
+++ b/R/plotutils.R
@@ -173,6 +173,11 @@ plotwitherror <- function(x, y, dy, ylim, dx, xlim, mdx, mdy, errsum.method="lin
     plot(x, y, col=col, ...)
   }
 
+  # We will need some langth in inches, so let's define the proportionality factor as
+  # size of the output in inches devided by the differnce of boundary coordinates.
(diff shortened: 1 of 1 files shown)


martin-ueding:
Nice, one can write “Resolves” and it will give the dependency between issue and pull request that I wanted (but did not look for). Great!

PR #149, opened 2019-02-03
Give not_na_by_design

### LINKED ISSUE #146: bootstrap.effectivemass should include which time indices are valid
@matfischer had a good idea to make interaction with his Bayesian fitting routines more straightforward.
In addition to ```t.idx```, ```bootstrap.effectivemass``` should produce a second vector of indices which can be used to subset the "valid" time slices of the effective mass, i.e., those which are not ```NA```. Since this depends on the type, it is good for ```boostrap.effectivemass``` to do this rather than having to externally check which effective mass type was used and making the appropriate externally.

martin-ueding:
Resolves #146.

Excuse my inability to name the branch correctly …


martin-ueding:
@matfischer has to decide what the function should do, so we have to halt this until then.

PR #151, opened 2019-02-03
Reduce opening windows to second plot

### LINKED ISSUE #138: plot.uwerr opens lots of windows by default
I created an `uwerr` object for each variable of the DAG-HMC project and then wanted to plot them all. Usually plots get caught in RStudio and then neatly displayed as a list of plots. With `plot.uwerr` I get a bunch of windows:

auswahl_001

Passing `x11 = FALSE` cures that problem, but I do not get why this is the default here while all the other plot functions seem to work without that. With that option the output is as expected:

auswahl_002

martin-ueding:
Resolves #138.

The X11 logic is refactored into a function and used everywhere. I have tested this for `plot.matrixfit`, it works both in RStudio and on the CLI. I think the trick is to not call it for the first plot.


kostrzewa:
So we should rename ```x11``` to ```interactive``` (or similar) and then use ```dev.new()``` instead of ```x11()``` to open new devices.

martin-ueding:
It seems that calling `dev.new()` does not do any harm in RStudio, so I think we can even call that before every plot in a context where we expect multiple ones.

kostrzewa:
@urbach can you test this please as it concerns mostly your workflow?

PR #153, opened 2019-02-05
Issue 96 refactor remove temporal

### LINKED ISSUE #96: removeTemporal.cf does too many things
On B55.32 we also see some thermal state (though perhaps just a constant) on the two-particle scattering:

auswahl_001

Perhaps physically this can only be a constant shift and trying to remove `M_\pi` from this would not make any sense. But if that makes sense, we cannot do that with `removeTemporal.cf` because it can only subtract the energy difference between two fitted states and here we only have one state.

I propose to refactor `removeTemporal.cf` into several smaller more focused functions:

- A dispersion relation, which we likely already have in hadron or LuescherAnalysis.
- Perhaps an utility which extracts the energy level and bootstrap samples from a fit.
- A weighting function which takes a correlation function, an energy and the energy bootstrap samples.
- Reduce `removeTemporal.cf` to take a correlation function, energy and bootstrap samples and let it call the weighting, the shift, and the weighting again.

martin-ueding:
I could not sleep, so I refactored `removeTemporal.cf` into a bunch of smaller non-exported functions and a just kept it as a driver function. This means that one can use `weight_shift_reweight.cf` to do just that with a given energy level. The old function is retained as `old_removeTemporal.cf` and there is a unit test comparing the results to each other.

Resolves #96.


martin-ueding:
PR #90 now somewhat depends on this as I would like to fix the things in the refactored version. But then the refactored version will differ from the old one. Could somebody please have a look that the test case proves that the refactoring was correct such that we can merge this?

maowerner:
The continuous integration test crashed with 
```
$ curl -fLo /tmp/pandoc-2.2-1-amd64.deb https://github.com/jgm/pandoc/releases/download/2.2/pandoc-2.2-1-amd64.deb
curl: (56) SSL read: error:00000000:lib(0):func(0):reason(0), errno 104
The command "curl -fLo /tmp/pandoc-2.2-1-amd64.deb https://github.com/jgm/pandoc/releases/download/2.2/pandoc-2.2-1-amd64.deb" failed and exited with 56 during .
```
Does someone know what could have caused this?

martin-ueding:
My idea is to first refactor this function and then extend it to `deltat != 1` in the other pull request.

The actual values used are not important as I test for exact numerical equality. Therefore I presume that just doing something is sufficient.

The Travis CI error message likely is a flaky internet connection on the cloud worker, happens all the time, just restart the job.

kostrzewa:
Is this still to be merged?

martin-ueding:
I think that for the three pion project I need something that works with more than just the removal of the energy difference of two pions. So I plan on getting back to that somewhat soon after I have fitted the pion correlation function and have the pion energy at hand.

urbach:
> I think that for the three pion project I need something that works with more than just the removal of the energy difference of two pions. So I plan on getting back to that somewhat soon after I have fitted the pion correlation function and have the pion energy at hand.

Why do you need more?

martin-ueding:
I just looked through this again. The old code was limited on `deltat = 1` and my new one is not fully there, but getting closer. Also it is more modular. I think that this can still be merged and that nothing regresses during that. So can we merge it?

kostrzewa:
As long as it doesn't break anything, please go ahead and pull it in,

PR #157, opened 2019-02-15
Eigenvalue sorting for t < t₀ (Issue #156)

### LINKED ISSUE #156: Eigenvalue sorting in GEVP wrong below t0
for t < t0 we have to reverse the sorting order, I think this is currently not done.

This might explain why principal correllator plots look as they do and also, why the pc model is not working as expected.

urbach:
can this be applied to the rho data for a test, please!?

this should provide a fix for #156


urbach:
could you try again?

urbach:
@martin-ueding could you please review lines 93 to 121 in `gevp.R` carefully? If this is now correct, I think we should check the impact on the rho data!

martin-ueding:
Yes, I'm on the review now.

martin-ueding:
I find this very peculiar:

auswahl_024

At (1) the `evectors` is initialized to be `NA`. And then in (2) it is used as a scalar product. This means that the result is all `NA` and one cannot get an ordering out of that. Is that intentional?

martin-ueding:
In following iterations that might not be a problem, but for the very first iteration it means that there are no reference.

urbach:
> In following iterations that might not be a problem, but for the very first iteration it means that there are no reference.

for `t==t0` always eigenvalue sorting is used, because no eigenvectors
are available... For the next t-value the evectors to compare with
should be available. Does this make sense?

martin-ueding:
Ah! I did not see the *or* clause in the preceeding `if`:

```r
if(sort.type == "values" || t == t0+1) {
```

I will check whether that is consistent now.

Another issue that I saw is that the scalar product is formed between *w* and *v* and not *v* and *v* as described in the documentation.

urbach:
> Another issue that I saw is that the scalar product is formed
> between *w* and *v* and not *v* and *v* as described in the
> documentation.

sorry, which scalar product do you mean?

martin-ueding:
This is the documentation for `gevp`:

docu

The scalar product should be done between the eigenvectors `v`, but in the code it is done between `w` and `v` actually. This is a paragraph from my review (not finished yet):

block

urbach:
okay, I see, well spotted! I have to think about it...

martin-ueding:
I have now finished my review: [gevp_review.pdf](https://github.com/HISKP-LQCD/hadron/files/2891223/gevp_review.pdf)

urbach:
thanks @martin-ueding ! Can we discuss this in person later?

PR #159, opened 2019-02-15
Avoiding Warnings

j-ostmeyer:


kostrzewa:
Agreed, this was a mistake. Could you also remove the comment above that line?

Issue #158, opened 2019-02-15
plotwitherror might omit error bars in log scale


NikSchlage:
Trying to plot the data in [strange_data.txt](https://github.com/HISKP-LQCD/hadron/files/2868770/strange_data.txt) using

```r
with(df, plotwitherror(x, y, dy, log = 'y'))
```

results in a plot where no error bars are shown:

![corr1t96l24](https://user-images.githubusercontent.com/41633300/52849704-f9ffb280-3111-11e9-85a7-0202b98be2a8.png)

Without the `log = 'y'` parameter, the error bars are shown.

With ggplot the error bars are visible in log-scale:

```r
ggplot(df, aes(x = x, y = y)) +
  geom_point() +
  geom_errorbar(aes(ymin = y - dy, ymax = y + dy)) +
  scale_y_log10() +
  theme_light()
```

![corr1t96l24_2](https://user-images.githubusercontent.com/41633300/52849960-9cb83100-3112-11e9-908f-ca066f89dd50.png)

So the data seems legitimate, but somehow `plotwitherror` does not show the errors.

kostrzewa:
This is due to fdc169c315fe7f651916728c21827bf8419c3c23 by @j-ostmeyer . The errors are so small that plotwitherror skips them. Incorrectly, it skips all of them. Reverting that commit makes things work as they should again.

kostrzewa:
Sorry, had the wrong commit id there...

kostrzewa:
@j-ostmeyer if you can find a better solution, that would be great. If not, I'm afraid we should probably revert your commit.

j-ostmeyer:
Thanks for the remark. I would not have thought that the error is there.

j-ostmeyer:
Solved by #160 .

kostrzewa:
@NikSchlage can you confirm that #160 solves this for you?

NikSchlage:
Yes, thank you all. Now the problem is solved and the error bars are shown:

![corr1_t96l24c](https://user-images.githubusercontent.com/41633300/52869291-a0b17680-3145-11e9-8ba2-4241c68a6882.png)

kostrzewa:
Okay, still looks a bit odd as a few error bars are missing, but I think we might just accept this, agreed @martin-ueding @urbach?

j-ostmeyer:
[...]

PR #160, opened 2019-02-15
Log scale arrows

j-ostmeyer:
Attempts to solve #158 .


kostrzewa:
This is tricky as now the width of the error bar heads, when the error bar itself cannot be resolved, are tied to the device width... While the length of error bars obviously should depend on the device coordinates (i.e., the physical size of the plot region). Just like the size of plot symbols, the width of error bars should be independent of the size of the plot region, but it should depend on the line height...

One automatic option would be to simply suppress warnings when drawing arrows.

j-ostmeyer:
> One automatic option would be to simply suppress warnings when drawing arrows.

True, but we explicitly wanted to avoid this.

kostrzewa:
Looks great

PR #163, opened 2019-02-22
Two features for nlsfit

j-ostmeyer:
Added a much faster version of error estimation. Instead of boostrap usually the error on fit results can be computed very precisely using the jacobian.
Added a flag for interpreting the errors as relative weights instead of absolute ones. This can be helpful for direct comparison with other fit routines.


urbach:
> Added a much faster version of error estimation. Instead of boostrap usually the error on fit results can be computed very precisely using the jacobian.

I'm actually not a big fan of this! If this is used with care in
exceptional cases, fine...

martin-ueding:
As far as I have understood Johann, he has tried this out for some of his cases and found good agreement. I have the impression that he wants to use it for special cases, bootstrap still being the default.

j-ostmeyer:
Precisely. The default has not changed. We just get an additional option that is faster and thus preferable for simple fit functions and tests. For the final and thorough version of an analysis we can (and should) still use the bootstrap.

urbach:
> Precisely. The default has not changed. We just get an additional
  option that is faster and thus preferable for simple fit functions
  and tests. For the final and thorough version of an analysis we can
  (and should) still use the bootstrap.

yes, I understand, of course.

urbach:
> > Precisely. The default has not changed. We just get an additional
>   option that is faster and thus preferable for simple fit functions
>   and tests. For the final and thorough version of an analysis we can
>   (and should) still use the bootstrap.
> 
> yes, I understand, of course.

what I don't understand yet: why does it need a new user interface
`simple.nlsfit`?

j-ostmeyer:
It doesn't. I just thought as we have it, we could as well provide it. But it has no real advantage. Shall I remove it from `NAMESPACE` again?

martin-ueding:
I have discussed that with Johann and I think that two separate functions make more sense with the [Interface Segregation Principle](https://en.wikipedia.org/wiki/Interface_segregation_principle).

urbach:
> It doesn't. I just thought as we have it, we could as well provide it. But it has no real advantage. Shall I remove it from `NAMESPACE` again?

well, I guess since it does not do bootstrap anymore, the original
name is missleading. So, better keep it.

martin-ueding:
What's the general status of this? The basic idea seems to be okay, can we merge it?

urbach:
This is fine with me. I didn't check the docu, though

j-ostmeyer:
I have been working with this version for three months now and did not have any issues. So I guess it's fine. Please check  and tell me if I should change anything, but I believe this is mergable.

PR #164, opened 2019-02-26
bootstrap.nlsfit: Accept info of 4 as convergence

martin-ueding:
With the IAM fit there is the `$info = 4` all the time. Currently we consider this to be a failure of the fit. From the documentation this means:

> The cosine of the angle between `fvec` and any column of the Jacobian is at most `gtol` in absolute value.

This reads as positive as reasons 1 and 2:

> Both actual and predicted relative reductions in the sum of squares are at most `ftol`.

And:

> Relative error between two consecutive iterates is at most `ptol`.

I have not really found something clearly saying which return values are to be considered good or bad. But I can generate an `$info = 4` with the fit of a constant (single parameter) to some data:

```{r}
y_val <- 17
y_boot <- rnorm(1000, 17, 1)

model <- function (par, x, ...) {
    par[1]
}

fit <- bootstrap.nlsfit(model, c(0), y_val, c(0), matrix(y_boot, ncol = 1))

plot(fit)
```

This works just fine, most fits have `$info = 2`, but one has `$info = 4`. Depending on the seed this will likely vary.

I have generated data from the IAM function with really low errors and started the fit directly with the known parameter, failed with `$info = 4`. Also fitting a Breit-Wigner to the same data gives `$info = 4`. 

When interpreting that as a success, the IAM fit to the artifical data converges even when starting from a different value of `\bar l`. Trying this on actual data (with fixed `\bar l` as guess) makes all the fits converge, both the steep and the shallow ones.

auswahl_011

auswahl_012

Now I suggest that we consider `$info = 4` a success as well.


martin-ueding:
Ah, now [this](https://www.rdocumentation.org/packages/minpack.lm/versions/1.2-1/topics/nls.lm) brings some light in:

> Third convergence test. This test is satisfied when the cosine of the angle between the result of `fn` evaluation `fvec` and any column of the Jacobian at `par` is at most `gtol` in absolute value. There is no clear relationship between this test and the accuracy of `nls.lm`, and furthermore, the test is equally well satisfied at other critical points, namely maximizers and saddle points. Therefore, termination caused by this test (`info = 4`) should be examined carefully. The recommended value for `gtol` is zero.

The default value in `minpack.lm::nls.lm.control` is `gtol = 0`. We do not specify that in `hadron::bootstrap.nlsfit`. Therefore we cannot make it any stricter.

But it also means that it is not clear-cut whether `info = 4` is converged or not. We likely want to allow the user to either specify something like `accept_info_4` (logical) or accept it but then give a `warning()` to the user. There is already `info.boot` included in the fit object, so one can investigate the return values.

martin-ueding:
I had disabled the X-errors to narrow down the error. In this case there is just one parameter and I would assume that the χ² is sufficiently well-behaved for this to actually give something other than a minimum. With X-errors included there are many more parameters and it might not look as well.

j-ostmeyer:
> But it also means that it is not clear-cut whether `info = 4` is converged or not.

This is why I am against including `info = 4` as converged by default.
But I agree that `info = 4` is much less severe than the other error infos, so we should introduce a possibility to distinguish there. I like the solution introducing a logical parameter with default `FALSE` allowing to include `info = 4` into the accepted values. If `TRUE` I would not give any warnings.

kostrzewa:
ping

martin-ueding:
I've just exposed `success_info = 1:3` as a function argument now, this way one can decide on a case-by-case basis.

martin-ueding:
I think that this is now viable and can be merged.

PR #165, opened 2019-03-05
Hotfix/cyprus_readutils

kostrzewa:
add support for "Naive" loop type and fix O(n) calculation of disconnected two point function


martin-ueding:
It does not affect me, so I have no objections.

PR #166, opened 2019-03-07
additions to work with loop and correlation function files generated …

kostrzewa:
…by CVC and stored in HDF5 files

This stuff is similar in spirit to what was done for ```cyprus_readutils```.


kostrzewa:
This is ready to go pending on merge of #200

kostrzewa:
Why did you undo the version bump? The analysis which relies on the functionality in this branch checks for version 3.3.2 for a reason...

martin-ueding:
I remove the bump because I find it cleaner to do so in the `master`. I had three pull requests which all wanted to change the version, which then ended up giving merge conflicts in the target version.

kostrzewa:
> I remove the bump because I find it cleaner to do so in the master. I had three pull requests which all wanted to change the version, which then ended up giving merge conflicts in the target version.

The merge conflicts are a necessary evil of this kind of concurrent development. It's a merge conflict in a single position and I believe that a feature branch should modify the version number rather than having to manually bump the version afterwards.

martin-ueding:
I like the way they do it in git-flow where you have a release branch and the version bump is done there. But you have a point. The problem is that when we have multiple long-running feature branches which both want to bump the version we will have different results depending on the order of their merging if one requests a minor and the other a major version bump. We can do either way, as long as the bump in the master branch is done properly and given a tag in git.

kostrzewa:
You have a point too with the tagging. I guess we do need to have a slightly different policy in that version bumps are done in master and tagged. On the flipside, this means that one can't rely on the version number when using a feature branch in production (which is the rule for us). Not much of a problem, but slightly annoying...

martin-ueding:
I see the issue with the tentative deployments. The version numbers are sequential, so basically you would be trying out the beta for the next feature. But for that one would have to land the feature in the beta-release branch and then bump the version there. I guess there is no real way to assign linear version numbers to a bouquet of feature branches. Also every other feature branch would have a bumped version number and therefore depending on the version alone is not sufficient in the most general case.

I try to get the branches merged such that I only have one feature branch in production and all the other new features already in `master`.

kostrzewa:
@martin-ueding I think you might just have nuked some of my commits with the force-push...

kostrzewa:
Yes, indeed you have...

kostrzewa:
I'll merge with master, remove my generated files and then force-push here

R/plotutils.R
```diff
@@ -159,7 +157,7 @@ errorpos <- function(dx,errsum.method="linear") {
 #' plotwitherror(x, y, dy)
 #' 
 #' @export plotwitherror
-plotwitherror <- function(x, y, dy, ylim, dx, xlim, mdx, mdy, errsum.method="linear.quadrature", rep=FALSE, col="black", ...) {
+plotwitherror <- function(x, y, dy, ylim = NULL, dx, xlim = NULL, mdx, mdy, errsum.method="linear.quadrature", rep=FALSE, col="black", ...) {
```
martin-ueding: What is the point of this? Isn't `missing` just fine?

kostrzewa: Basically this is a convenience for client code. Imagine you have branching in an analysis code which computes something for `ylim` if some condition(s) is(are) met or delegates its calculation to the algorithm within `plotwitherror` if the condition(s) is(are) not met.

Using `missing()`, the branching needs to produce two different calls of `plotwitherror`. On the other hand, when `NULL` is used instead, you have a single call and just modify whether you pass `NULL` or not.

kostrzewa: This reminds me to review the `plotwitherror` documentation one of these days because it's not fully representative of the current state of the function.

kostrzewa: Of course, one could achieve something similar using `do.call` and having the branching produce different argument lists.

kostrzewa:
Well, if you agree, I'd change this to be `3.0.0` now that the dust has settled.

R/plotutils.R
martin-ueding: Indeed, using `NULL` makes this more convenient for the caller.

martin-ueding:
We can do 3.0.0. The installation process has changed, so a major bump feel good for that.

PR #169, opened 2019-05-11
Rename matrixfit functions

martin-ueding:
We now have `new_matrixfit` around for long enough, we should make the switch to the new function.

Blockers for merge:

- [ ] default ```optim``` codepath should work
- [ ] confidence intervals should be displayed with more digits

Desirable (but not essential) improvements:

- [ ] time to solution parity with old version (but perhaps the new version is more reliable and worth the cost?)


kostrzewa:
I guess this should also include a rename of the files.

kostrzewa:
Was this checked to work with larger matrices with mixed sym.vec ?

kostrzewa:
I guess the interface is unchanged, correct?

kostrzewa:
ping

kostrzewa:
Was this checked to work with larger matrices with mixed sym.vec ?

martin-ueding:
I think that we have never tested this as I have just fitted principal correlators up to this point.

martin-ueding:
I think that the existing models should continue to work as before because the `sign.vec` and `ov.sign.vec` are passed as before, the model code has been copied. Given some data we could add another unit test, but I would think that this is good enough to be merged. I'd need a review by somebody else to merge it in.

kostrzewa:
Right away this breaks my fits, so there must have been some changes to the interface...

kostrzewa:
Broken call:

```
      mfit <- hadron::matrixfit(cf = value$cf_2pt_conn$data[[sel$data_idx]],
                                t1 = value$t1,
                                t2 = value$t2,
                                useCov = TRUE,
                                autoproceed = FALSE)
```

kostrzewa:
Error traceback is both more useful (as it contains more information) and less useful (as it contains way too much information):

```
Error in fn(par, ...) : unused argument (boot_r = 0)
In addition: Warning messages:
1: In hadron::matrixfit(cf = value$cf_2pt_conn$data[[sel$data_idx]],  :
  missing sym.vec, using default for single correlator!
2: In hadron::matrixfit(cf = value$cf_2pt_conn$data[[sel$data_idx]],  :
  missing neg.vec, using default (correlator positive)!
> traceback()
17: (function (par) 
    fn(par, ...))(c(0.168754261667395, 0.467569297859208))
16: optim(par = par, fn = fitchisqr, gr = dfitchisqr, y = y, method = c("BFGS"), 
        control = list(maxit = maxiter), ...)
15: wrapper(Y, par.Guess, boot_r = 0, ...)
14: (function (fn, par.guess, y, x, bsamples, ..., dy, dx, CovMatrix, 
        gr, dfn, use.minpack.lm = TRUE, parallel = FALSE, error = sd, 
        cov_fn = cov, maxiter = 500, relative.weights = FALSE) 
    {
        stopifnot(!missing(y))
        stopifnot(!missing(x))
        stopifnot(!missing(par.guess))
        stopifnot(!missing(fn))
        stopifnot(!missing(bsamples))
        boot.R <- nrow(bsamples)
        useCov <- !missing(CovMatrix)
[...]
```

lots of code snippets and data beyond that, until the calling line in my analysis is found.

kostrzewa:
I think we encountered this before, it seems that the call too ```optim``` is broken. I'll try the ```lm``` call.

kostrzewa:
These changes have also slowed down matrixfit by about a factor of three!

kostrzewa:
Using ```fit.method = lm```, the call works, but the slowdown is quite severe!

**old matrixfit (using ```fit.method = lm```):**

```
# A tibble: 1 x 6
  lattice_spacing_letter ensemble_base ensemble    cor_id   src_psq snk_psq
  <chr>                  <chr>         <chr>       <chr>      <int>   <int>
1 A                      cA211.53      cA211.53.24 pipm_2pt       0       0
values:  cf_2pt_conn t1 t2 extent_space extent_time mu_l_sea mu_sigma_sea mu_delta_sea kappa_sea 
  |++++++++++++++++++++++++++++++++++++++++++++++++++| 100% elapsed = 11s
# A tibble: 1 x 6
  lattice_spacing_letter ensemble_base ensemble    cor_id   src_psq snk_psq
  <chr>                  <chr>         <chr>       <chr>      <int>   <int>
1 A                      cA211.30      cA211.30.32 pipm_2pt       0       0
values:  cf_2pt_conn t1 t2 extent_space extent_time mu_l_sea mu_sigma_sea mu_delta_sea kappa_sea 
  |++++++++++++++++++++++++++++++++++++++++++++++++++| 100% elapsed = 11s
Saving fits_2pt_fns ... took 0.39 seconds.
There were 50 or more warnings (use warnings() to see the first 50)
```

** new matrixfit (using ```fit.method = lm```): **

```
# A tibble: 1 x 6
  lattice_spacing_letter ensemble_base ensemble    cor_id   src_psq snk_psq
  <chr>                  <chr>         <chr>       <chr>      <int>   <int>
1 A                      cA211.53      cA211.53.24 pipm_2pt       0       0
values:  cf_2pt_conn t1 t2 extent_space extent_time mu_l_sea mu_sigma_sea mu_delta_sea kappa_sea 
  |++++++++++++++++++++++++++++++++++++++++++++++++++| 100% elapsed = 27s
# A tibble: 1 x 6
  lattice_spacing_letter ensemble_base ensemble    cor_id   src_psq snk_psq
  <chr>                  <chr>         <chr>       <chr>      <int>   <int>
1 A                      cA211.30      cA211.30.32 pipm_2pt       0       0
values:  cf_2pt_conn t1 t2 extent_space extent_time mu_l_sea mu_sigma_sea mu_delta_sea kappa_sea 
  |++++++++++++++++++++++++++++++++++++++++++++++++++| 100% elapsed = 25s
Saving fits_2pt_fns ... took 0.32 seconds.
```

kostrzewa:
Even using ```optim```, the old code is faster...

kostrzewa:
I really appreciate how much cleaner this is (although it is also more complicated in some sense), but it would be nice to understand why it has become so much slower. In addition, the ```optim``` (default!) codepath should work...

R/matrixfit.R
```diff
@@ -1,791 +1,462 @@
...
-deriv.CExp.shifted <- function(par, t, T, sign, deltat=1) {
-  res <- array(0.,dim=c(length(par),length(t)))
-  
-  res[1,] <- par[2]*par[3]*(-(t-deltat/2)*exp(-par[1]*(t-deltat/2)) + (T-(t-deltat/2))*sign*exp(-par[1]*(T-(t-deltat/2))))
-  res[2,] <- par[3]*(exp(-par[1]*(t-deltat/2)) - sign*exp(-par[1]*(T-(t-deltat/2))))
-  res[3,] <- par[2]*(exp(-par[1]*(t-deltat/2)) - sign*exp(-par[1]*(T-(t-deltat/2))))
-
-  return(res)
-}
+library(R6)
```
kostrzewa: is the usage of ```library()``` in a package legitimate for ```R6```? Weren't we supposed to use ```requireNameSpace```?

R/matrixfit.R
```diff
@@ -1,791 +1,462 @@
...
+
+MatrixModel <- R6Class(
```
kostrzewa: what's the policy on documentation for R6 classes?

kostrzewa:
The generalisation of the summary is a bit "meh", but I guess there's not much that can be done about it.

```
          par      bias ci16 ci84
1 0.16661(86) -0.000014 0.17 0.17
2  0.4606(34)  0.000011 0.46 0.46

   correlation matrix of the fit parameters

         X1        X2
1 1.0000000 0.6570536
2 0.6570536 1.0000000

   chi^2 and fit quality
chisqr / dof = 22.07061 / 10 = 2.207061 
p-value 0.01474888 

Ratio of converged fits on samples: 1500 / 1500 = 1 
Table of nls.lm info values (1, 2, 3 are convergence):
  value frequency
1     1      1487
2     2         6
3     3         7
  value frequency
1     1      1487
2     2         6
3     3         7
```

For the output of the ```ci16 / ci84``` to be useful, more digits should be shown.

kostrzewa:
In the summary, ```boot.R, boot.l, useCov``` and some more pertinent information should be displayed

R/matrixfit.R
martin-ueding: Yes, because it is a compile time dependency at this point. We need to to declare the class, but we do not need it at runtime.

martin-ueding:
I have used `profvis` to do a profile of the old `matrixfit` at `master`:

Auswahl_027

Now the `master` version of the new `matrixfit` looks like this:

Auswahl_028

In this branch the old `matrixfit` is a bit different, I do not fully get why:

Auswahl_029

And the new `matrixfit` in this branch:

Auswahl_030

Perhaps these are just fluctuations due to speed scaling on my laptop. But the general trend is still visible. Both seem to spend most of their time in `minpack.lm::nls.lm`. These are my guesses:

- The profiles are wrong because the code only runs for half a second.

- The new `matrixfit` using `bootstrap.nlsfit` uses different starting parameter for the fits to the samples and therefore it just takes more iterations to converge.

- Due to more layers of virtual function calls we spend more time with overheads for computing the χ vector. This then makes it more expensive inside of `nls.lm`.

For 1500 bootstrap samples the old code spends accumulated 140 ms in the function and 380 ms in the gradient from 740 ms total. The new code spends 360 ms in the function and 780 ms in the gradient from 1330 ms total.

martin-ueding:
I have tried using 15000 bootstrap samples instead of 1500 next. This way we run longer and should have more reliable estimates.


Code | Function | Gradient | Total
--- | ---: | ---: | ---:
Old | 960 ms | 3210 ms | 6620 ms
New | 2450 ms | 9720 ms | 14170 ms

Unfortunately it does not tell me the number of function calls. My plan would be to go through the code and see how many calls to `nls.lm` are really made. I think the old code does 1 + R calls whereas the new code does 2 + R calls. Then I will look at the number of iterations that we have needed in each of them to figure out how many calls to the function and gradient we actually dispatched to figure out whether that is the culprit or the additional abstractions.

kostrzewa:
> The new matrixfit using bootstrap.nlsfit uses different starting parameter for the fits to the samples and therefore it just takes more iterations to converge.

I would expect that this will turn out to be the reason.

martin-ueding:
So I have extracted the number of iterations for both cases and it seems that the fit converges super fast in both cases when using `nls.lm`:

Auswahl_002

So I would think that this is not the issue, or I just extracted the wrong information in both cases.

martin-ueding:
It seems that OO virtual function calls are quite expensive to [Hadley Wickham's “Advanced R” chapter](http://adv-r.had.co.nz/Performance.html). Since the number of iterations has not significantly changed and from the profile we gather that `nls.lm` takes the most time, I would conclude that it is the R6 “vtable” resolution overhead. This is amplified by another layer of closures that `bootstrap.nlsfit` puts over it. The old `matrixfit` just directly uses chi-vector and the Jacobian as function pointers, the new `matrxfit` has R6 classes with the model and gradient, then `bootstrap.nlsfit` add closures for chi-vector and Jacobian from that. And the actual computation of the chi for five time slices is pretty cheap and likely using compiled code deep in R, so we are likely up against the overheads here.

martin-ueding:
For the output with `summary` we currently use `summary.bootstrapfit`. This means that there is no information about the `cf` object that we had put in. The output of `summary.matrixfit` looks as follows:

```
 ** Result of one state exponential fit **

based on 1018 measurements
time range from 5  to  10 

ground state energy:
E 	=	 0.1488968 
dE	=	 0.0002465713 

Amplitudes:
P 1 	=	 25.65402 
dP 1 	=	 0.02775326 

boot.R	=	 1500  (bootstrap samples)
boot.l	=	 2  (block length)
useCov	=	 FALSE 
chisqr	=	 5.747993 
dof	=	 4 
chisqr/dof=	 1.436998 
Quality of the fit (p-value): 0.2187753 
```

Whereas the output of `summary.bootstrapfit` just looks like this:

```
bootstrap nls fit

model yerrors 
    best fit parameters with errors, bootstrap bias and 68% confidence interval

          par      bias ci16 ci84
1 0.14890(25) 0.0000025 0.15 0.15
2  25.654(28)  -0.00069   26   26

   correlation matrix of the fit parameters

         X1        X2
1  1.000000 -0.166182
2 -0.166182  1.000000

   chi^2 and fit quality
chisqr / dof = 5.747993 / 4 = 1.436998 
p-value 0.2187753 

Ratio of converged fits on samples: 1500 / 1500 = 1 
Table of nls.lm info values (1, 2, 3 are convergence):
  value frequency
1     1       823
2     2       130
3     3       547
```

There is information that the general `summary.bootstrapfit` just does not know at this point, like the meaning of the fit parameters and the number of measurements that are contained in the correlation function. Additionally `plot.bootstrapfit` only plots the time slices that were actually fitted, and not the other ones.

I see two options to work on this:

1. Add a class `new_matrixfit` (or rename the other to `old_matrixfit` and have the new one `matrixfit`). Add the `cf` object to the fit object. Have `summary.new_matrixfit` and a `plot.new_matrixfit` which resemble the output of the old functions.

2. Add more general parameters to `bootstrap.nlsfit` which take the names for the fit parameters and additional meta data that can be put into the summary.

Actually I think that renaming existing things while keeping them around is a lot of work. Perhaps we should just call the new one `matrixfit2`?

kostrzewa:
Unfortunately, this also breaks existing analyses by removing ```opt.res``` and ```opt.tsboot```. I don't think that that's necessarily a bad thing, because the change clearly moves everything towards being more consistent, but I think @urbach should take a look at this and make the final decision on how to proceed...

urbach:
> Unfortunately, this also breaks existing analyses by removing ```opt.res``` and ```opt.tsboot```. I don't think that that's necessarily a bad thing, because the change clearly moves everything towards being more consistent, but I think @urbach should take a look at this and make the final decision on how to proceed...

In my opinion we should go ahead!

kostrzewa:
> In my opinion we should go ahead!

Go ahead with what?

urbach:
> > In my opinion we should go ahead!
> 
> Go ahead with what?

it okay that changes break older code. I think more consistency is
important here.

kostrzewa:
> it okay that changes break older code. I think more consistency is important here.

Agreed. What about the performance regression noted above? This has not been tested on actual correlator matrices so there's no telling how the regression scales. What if it scales with the square of the matrix size and the new code is suddenly two orders of magnitude slower than the old one just because of indirection overheads?

urbach:
> > it okay that changes break older code. I think more consistency is important here.
> 
> Agreed. What about the performance regression noted above? This has not been tested on actual correlator matrices so there's no telling how the regression scales. What if it scales with the square of the matrix size and the new code is suddenly two orders of magnitude slower than the old one just because of indirection overheads?

that's going to be a big problem, needs testing

kostrzewa:
> that's going to be a big problem, needs testing

do you happen to have data lying around (any format, we can always convert) for a correlator matrix with mixed time reflection? This is the most general case which should be included as a basic test.

martin-ueding:
I now think that we should not rename these functions but instead give them a new name, perhaps `matrixfit_v2` instead of `new_matrixfit` because this way we could do another refactoring without breaking things.

People using the function could opt-in to the new code at any time that would be convenient for them. This would remove most of the pains from renaming and does not introduce performance regressions without users explicitly opting in.

kostrzewa:
> > that's going to be a big problem, needs testing
> 
> do you happen to have data lying around (any format, we can always convert) for a correlator matrix with mixed time reflection? This is the most general case which should be included as a basic test.

@curbach ?

martin-ueding:
> do you happen to have data lying around (any format, we can always convert) for a correlator matrix with mixed time reflection?

Don't we have such correlators in the rho project?

PR #170, opened 2019-05-14
Fix txt bar

urbach:
when `readcmidatafiles` is used with a single file the `txtProgressBar` fails because min is not smaller than max. This is fixed here by setting `min=0`

### DIFF
--- a/R/readutils.R
+++ b/R/readutils.R
@@ -675,7 +675,7 @@ readgradflow <- function(path, skip=0, basename="gradflow", col.names) {
   rm(tmpdata)
   
   pb <- NULL
-  pb <- txtProgressBar(min = 1, max = length(files), style = 3)
+  pb <- txtProgressBar(min = 0, max = length(files), style = 3)
   for( i in 1:length(files) ){
     setTxtProgressBar(pb, i)
     tmp <- data.frame()
--- a/R/readutils.R
+++ b/R/readutils.R
@@ -81,7 +81,7 @@ readcmifiles <- function(files, excludelist=c(""), skip, verbose=FALSE,
   
   pb <- NULL
   if(!verbose) {
-    pb <- txtProgressBar(min = 1, max = nFiles, style = 3)
+    pb <- txtProgressBar(min = 0, max = nFiles, style = 3)
   }
   for(i in c(1:nFiles)) {
     # update progress bar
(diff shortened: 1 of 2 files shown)


kostrzewa:
@martin-ueding can you confirm that you agree with the change in tolerance in the test?

martin-ueding:
I have just tested the `master` and your `fixTxtBar` branch (with `1e-5` and `1e-6`) locally with my R 3.5.3 on Fedora 30. It works for both tolerances. I would not have understood how your chance could change the tests.

It seems that it just violates the tolerance just by a tiny bit on Travis CI:

    [2] 0.798 - 0.798 == 1.04e-06

On Travis CI we now use R 3.6.0, which was released a few weeks ago. The latest `master` compiled fine on Travis CI, but that commit was made on 2019-03-12, about a month before the new release of R. Suspecting that the new version of R broke our test, I have restarted the run of `master` on Travis CI. And that also fails with exactly the same number. The issue therefore is not your pull request but the new version of R (perhaps only on Travis CI).

Assuming that the R developers did not break anything fundamental, I would just assume that the random number generator has changed and therefore the bootstrap procedure is slightly different. Our seed is fixed by the default argument value in `bootstrap.cf`, so that cannot be it. But perhaps the RNG was changed?

Looking at the [release notes of 3.6.0](https://stat.ethz.ch/pipermail/r-announce/2019/000641.html) supports my hunch:

> The default method for generating from a discrete uniform distribution (used in `sample()`, for instance) has been changed. This addresses the fact, pointed out by Ottoboni and Stark, that the previous method made `sample()` noticeably non-uniform on large populations.  See `PR#17494` for a discussion.  The previous method can be requested using `RNGkind()` or `RNGversion()` if necessary for reproduction of old results.  Thanks to Duncan Murdoch for contributing the patch and Gabe Becker for further assistance.
>
> The output of `RNGkind()` has been changed to also return the 'kind' used by `sample()`.

Therefore I would conclude that this is indeed a slight difference in random sampling. In the other tests we also only have a tolerance of just `1e-4`, so relaxing to `1e-5` is fine. So I'm fine with relaxing this to `1e-5`.

kostrzewa:
It's absolutely clear that it's not @urbach's change which caused this, the same regression appeared in other travis runs already. I would like to make sure, however, that we understand why the fit result changed to an extent that made this test fail.

The failure is in the value of the fit parameter (and not its error), so the only thing that can affect this is that the variance-covariance matrix estimated from the bootstrap samples (if indeed used here) changed sufficiently to cause the fit to converge to a different value at the sixth digit (we don't know to what extent the tolerance was satisfied previously, so we have to assume worst case). Or do you use some random data for this test? (haven't read the test case in detail)

What's the statistical uncertainty on the value that is being compared in the test?

PR #176, opened 2019-08-06
Closer to cran

martin-ueding:
I have made a few more tweaks in order to remove more warnings. We are not there yet, but now the main thing missing is documentation of functions that I do not know.

Also I think it would not be clever for me to submit but Carsten should do that in order to become the owner of that package. We then should do the same with rzeta and paramvalf such that students can just install them all from CRAN.


NAMESPACE
```diff
@@ -47,3 +47,53 @@ importFrom("stats", "acf", "approx", "complete.cases", "cov", "deltat",
...
+S3method(summary, effectivemass)
+S3method(summary, effectivemassfit)
+S3method(summary, gevp.amplitude)
+S3method(summary, hadronacf)
+S3method(summary, matrixfit)
+S3method(summary, ofit)
+S3method(summary, pionff)
+S3method(summary, pionfit)
+S3method(summary, raw_cf)
+S3method(summary, uwerr)
```
urbach: why is this needed?

R/raw_cf.R
```diff
@@ -557,7 +557,7 @@ get_plotdata_raw_cf <- function(cf,
 #' @param reim_same Boolean, determines whether the real and imaginary parts, if both
 #'                  are to be plotted, will be plotted in the same plot.
 #' @param ... Further parameters passed to \link{plotwitherror}.
-#' @return Invisibly returns the plotdata, see \link{get_plotdata.raw_cf}.
+#' @return Invisibly returns the plotdata, see \link{get_plotdata_raw_cf}.
```
urbach: '.' here, but not in `plot.raw_cf` and others?

kostrzewa: The choice between `.` and `_` is based on whether a generic function exists or not not. Since `get_plotdata_raw_cf` is a utility function (which is probably not even exported), it makes sense for it to not use the dot notation. On the other hand, `plot.raw_cf` can be dispatched by the `plot` generic function. Is that what you meant?

kostrzewa: Things are not quite consistent, however, as exemplified by `uwerr`...

kostrzewa: And there are some generic functions missing, for which it might actually be interesting to provide prototypes.

urbach:
do we need in your opinion to export the following functions:

- compute.plotlims
- compute.qtildesq
- compute.qtildesq.contdisp
- effmass2
- g1
- pcacfit
- read_nissa_textcf_kernel
- symmetrise.cf
- uwerr.cf

kostrzewa:
> * read_nissa_textcf_kernel

no, that's a mistake


> * symmetrise.cf

yes

> * uwerr.cf

yes

urbach:
so, the docu is fixed and I merged to master. We could merge this in, I think it has no consequences other than more docu!?

R/analysis_online.R
```diff
@@ -15,6 +15,66 @@ append_pdf_filename <- function(basename, pdf_filenames){
...
+#' 
+#' @param L integer. spatial lattice extend
+#' @param T integer. temporal lattice extend
+#' @param t1 integer. initial time of fit range
+#' @param t2 integer. end time of fit range
+#' @param beta numeric. inverse squared gauge coupling
+#' @param kappa numeric. hopping parameter
+#' @param mul numeric. light twisted mass
+#' @param cg_col integer. column of CG iteration counts to use
+#' @param evals ??. ??
```
urbach: @kostrzewa could you provide a replacement for ??

R/analysis_online.R
```diff
@@ -15,6 +15,66 @@ append_pdf_filename <- function(basename, pdf_filenames){
...
+#' @param oldnorm boolean. If `TRUE` use the old tmLQCD normalisation
+#' @param S numeric. `S` parameter of \link{uwerr}
+#' @param stat_skip integer. statistics to skip
+#' @param omeas.samples integer. number of stochastic sample per online measurement
+#' @param omeas.stride integer. use as strides in online measurements
+#' @param omeas.avg integer. ??
+#' @param omeas.start integer. ??
+#' @param omeas.stepsize integer. ??
+#' @param evals.start integer. ??
+#' @param evals.stepsize integer. ??
```
urbach: @kostrzewa also here?

R/raw_cf.R
urbach: okay, I got it

NAMESPACE
martin-ueding: When one does not attaches the namespace, the S3 method resolution does not work:

```
> plot(hadron::samplecf)
Fehler in xy.coords(x, y, xlabel, ylabel, log) : 
  'x' is a list, but does not have components 'x' and 'y'
```

With these explicit S3 methods, the above should work, but I have not tested it.

R/analysis_online.R
kostrzewa: done

kostrzewa:
Man, `analysis_online` is such a steaming pile of ...

kostrzewa:
Running checks on the various additions now.

kostrzewa:
We're very close now, just a few exported objects remaining to be documented.

urbach:
> Man, `analysis_online` is such a steaming pile of ...

yep!

kostrzewa:
> Man, `analysis_online` is such a steaming pile of ...
> yep!

I'd really like to modularise it and add support for multi-replica analysis, but that will probably take several days which are simply not on the horizon right now...

urbach:
do we want to export

‘weight.cf’ and ‘weight_shift_reweight.cf’

?

martin-ueding:
Yes please, I use them in the LüscherAnalysis package.

urbach:
> Yes please, I use them in the LüscherAnalysis package.

I think you wrote them, so, could you maybe, if it doesn't create too
much of a problem ... write the docu?!

urbach:
see pull request #198

urbach:
is `devtools:load_all()` in `jackknife_cov_and_missing_values.Rmd`
really needed?

urbach:
and given this here from R CMD check:

gevp2cf: possible error in cf_weighted(cf, weight.factor =
  gevp$cf$weight.factor, weight.cosh = gevp$cf$weight.cosh, mass1 =
  gevp$cf$mass1, mass2 = gevp$cf$mass2): unused arguments (mass1 =
  gevp$cf$mass1, mass2 = gevp$cf$mass2)

Maybe we should still merge the fix from @ueding in?

martin-ueding:
And yes, we should merge the fix from me there, I'll see about the failing unit tests. In the current form the function is just wrong for `cosh.factor != 0.0`.

urbach:
one thing left:

    Codoc mismatches from documentation object 'weight.cf':
    weight.cf
      Code: function(cf, energy_difference_val, energy_difference_boot,
                     cosh_factor, offset = 0, inverse = FALSE)
      Docs: function(cf, energy_difference_val, energy_difference_boot,
                     cosh_factor, offset = 0)
      Argument names in code not in docs:
        inverse

urbach:
it was still the old docu in the repo...

martin-ueding:
I believe that this was because the function has been changed in a different branch. I'll fix it.

PR #184, opened 2019-09-27
Fix various issues in removeTemporal

martin-ueding:


urbach:
why does the test still fail?

urbach:
Is this working already with `deltat` not equal to 1?

martin-ueding:
The test now passes locally. I had to modify the sign in the old function (at the top of the file). So please have a look whether you agree with the fix there. Then the new function should be fine because it reproduces the unit tests.

martin-ueding:
I believe that `deltat` ≠ 1 does not work yet, but we are closer to it than before. Could we merge this, such that #193 also advances?

PR #186, opened 2019-10-07
add more first-class support for the imaginary part of the cf

kostrzewa:
It's nice to be able to count on the imaginary part of the ```cf``` to be taken care of during standard manipulations.


kostrzewa:
Of course, there is something somewhat important missing: there's no scaling by a complex constant.

R/cf.R
```diff
@@ -69,14 +70,16 @@ cf_meta <- function (.cf = cf(), nrObs = 1, Time = NA, nrStypes = 1, symmetrised
 #' @family cf constructors
 #'
 #' @export
-cf_boot <- function (.cf = cf(), boot.R, boot.l, seed, sim, cf.tsboot, resampling_method = 'bootstrap') {
+cf_boot <- function (.cf = cf(), boot.R, boot.l, seed, sim, cf.tsboot, icf.tsboot = NULL, resampling_method) {
```
kostrzewa: For safety reasons, I would like to force the specification of ```resampling_method```

R/cf.R
```diff
@@ -861,11 +941,20 @@ summary.cf <- function(object, ...) {
...
+      out <- cbind(iC=cf$icf0, itsboot.se = cf$itsboot.se)
+    }
   }
 
   if (inherits(cf, 'cf_jackknife')) {
     out <- cbind(out, jackknife.se=cf$jackknife.se)
     out <- cbind(out, jab.se=cf$jack.boot.se)
+
+    if( !is.null(cf$icf) ){
+      out <- cbind(out, ijackknife.se = cf$ijackknife.se)
```
kostrzewa: I haven't touched ```jackknifeafterboot.R``` yet

kostrzewa:
Some stuff will break as I move along here...

kostrzewa:
@martin-ueding A change which might affect your analyses / tests is that I've set ```cf$icf <- NULL``` in the sample data explicitly now (rather than arrays of ```NA```)

R/cf.R
```diff
@@ -585,6 +606,9 @@ add.cf <- function(cf1, cf2, a = 1.0, b = 1.0) {
 
 #' Arithmetically divide correlators
```
kostrzewa: Division does not make too much sense anyway, so I'm not too worried that what is done below is clearly not correct. For the cases where the signal of interest is either in the real or imaginary part, it would do the correct thing.

martin-ueding:
I think that she should have a definition of the `cf_orig` such that one can properly test for `icf`. So either it is always `NULL` (equivalent to not defined at all) or set to a scalar `NA` or to an array of NA of the same size as `cf`. I think that the first one indeed makes the most sense, and then one should act accordingly.

Perhaps we also just add a `has_icf <- function (cf) !is.null(cf$icf)` to make this implementation independent.

kostrzewa:
> I think that she should have a definition of the cf_orig such that one can properly test for icf. So either it is always NULL (equivalent to not defined at all) 

What's insufficient in the current implementation in your opionion?

```
cf_orig <- function (.cf = cf(), cf, icf = NULL) {
  stopifnot(inherits(.cf, 'cf'))

  .cf$cf <- cf
  .cf$icf <- icf

  class(.cf) <- append(class(.cf), 'cf_orig')
  return (.cf)
}
```

> Perhaps we also just add a has_icf <- function (cf) !is.null(cf$icf) to make this implementation independent.

Excellent plan, will do.

martin-ueding:
> What's insufficient in the current implementation in your opinion?

The implementation is fine and it serves as substitute for a formal definition of the class. So I think that this is just as fine as one can get with an S3 class.

R/cf.R
```diff
@@ -372,24 +383,28 @@ bootstrap.cf <- function(cf, boot.R=400, boot.l=2, seed=1234, sim="geom", endcor
   stopifnot(boot.l <= nrow(cf$cf))
   stopifnot(boot.R >= 1)
 
-  ## save random number generator state
```
kostrzewa: removed this as it's not required due to the usage of `swap_seed` below

R/cf.R
kostrzewa: partial support now available

kostrzewa:
I would like to skip `jab.cf.derived` as it's not quite clear what kind of derived quantity can be computed from the real and imaginary parts together. With that said, this is ready for review.

kostrzewa:
On a side-note, the error message there is confusing, isn't it @urbach?:

```r
jab.matrixfit <- function(cf, m=1) {
  if(!any(class(cf) == "matrixfit")) {
    stop("bootstrap.cf requires an object of class 'matrixfit' as input! Aborting!\n")
  }
  cf$jack.boot.se <- jab.cf.derived(cf=cf, m=m)
  return(invisible(cf))
}
```

kostrzewa:
@urbach should we address the point above in jackknifeafterbootstrap or should we just merge this one line change?

kostrzewa:
Alright, since there are no responses I will just merge this last bit in...

PR #197, opened 2019-11-13
sort of inverse function for tex.catwitherror

urbach:
might be useful...?


martin-ueding:
Yes, this can be helpful! I have added a few unit tests to make sure that it works for some edge cases, and some of them failed:

```
test_string2error.R:5: failure: 1
string2error("-3.10(2)") not equal to c(-3.1, 0.02).
1/2 mismatches
[2] -0.02 - 0.02 == -0.04

test_string2error.R:8: failure: 1
string2error("-310(2)") not equal to c(-310, 2).
1/2 mismatches
[2] -2 - 2 == -4

test_string2error.R:13: failure: 1
string2error("0.001(2000)") not equal to c(0.001, 2).
1/2 mismatches
[2] 2000 - 2 == 1998

test_string2error.R:14: failure: 1
string2error("0.001(20000)") not equal to c(0.001, 20).
1/2 mismatches
[2] 20000 - 20 == 19980
```

The error also comes out with a minus sign when the value is negative. Also when the error is so large that it spills over the decimal point, the function does not give the correct result.

I have implemented this a bit differently now and all my test cases pass. But perhaps there are more edge cases that I missed?

urbach:
> martin-ueding approved this pull request.
> 
> I'm basically reviewing my own code now. As I have backed it up with unit tests, I still think that it is somewhat fair.

I think the sign of the error should be positive. Otherwise I'm fine
with your re-writing.

martin-ueding:
I believe that the sign of the value has been inherited to the error from the string concatenation. Now with the regular expression matching this cannot happen. And if a user would specify something like `1.23(-45)`, it would not be matched and therefore be rejected.

urbach:
okay, agreed!

tests/testthat/test_string2error.R
```diff
@@ -0,0 +1,16 @@
...
+  
+  expect_equal(string2error('310(2)'), c(310, 2))
+  expect_equal(string2error('-310(2)'), c(-310, 2))
+  
+  expect_equal(string2error('0.001(2)'), c(0.001, 0.002))
+  expect_equal(string2error('0.001(20)'), c(0.001, 0.020))
+  expect_equal(string2error('0.001(200)'), c(0.001, 0.200))
+  expect_equal(string2error('0.001(2000)'), c(0.001, 2.000))
+  expect_equal(string2error('0.001(20000)'), c(0.001, 20.000))
+})
```
urbach: add newline?

PR #198, opened 2019-11-14
some final docu problems fixed

urbach:
some final things in the docu towards CRAN


urbach:
thanks!

kostrzewa:
Ready to merge?

R/removeTemporal.cf.R
```diff
@@ -180,7 +180,7 @@ old_removeTemporal.cf <- function(cf,
 #' @return
 #' The shifted correlator as an object of type `cf`, see \link{cf}
 #'
-#' @eport
+#' @export
```
martin-ueding: This is of course a needed fix. But keep in mind that it does not have any effect at this point because of our hand-written `NAMESPACE` file.

urbach:
> martin-ueding approved this pull request.
> 
> 
> 
> > -#' @eport
> +#' @export
> 
> This is of course a needed fix. But keep in mind that it does not have any effect at this point because of our hand-written `NAMESPACE` file.

yes, I know...

kostrzewa:
Pressing the button in 60 seconds

kostrzewa:
Alright, I'll hold off as there seem to be more commits

urbach:
should be fine now!

urbach:
so, how do we do this now...!?

urbach:
left in `R CMD check` is now only this and the R6 stuff:

    * checking R code for possible problems ... NOTE
    analysis_online: no visible binding for global variable ‘evals’
    cyprus_read_loops: no visible global function definition for ‘h5ls’
    cyprus_read_loops: no visible global function definition for
      ‘H5Lexists’
    cyprus_read_loops: no visible global function definition for ‘H5Fclose’
    gevp2cf: possible error in cf_weighted(cf, weight.factor =
      gevp$cf$weight.factor, weight.cosh = gevp$cf$weight.cosh, mass1 =
      gevp$cf$mass1, mass2 = gevp$cf$mass2): unused arguments (mass1 =
      gevp$cf$mass1, mass2 = gevp$cf$mass2)
    loop_2pt: no visible binding for global variable ‘r1’
    loop_2pt: no visible binding for global variable ‘r2’
    set.dfitchi : jacobian: no visible binding for global variable ‘nx’
    set.dfitchi : jacobian: no visible binding for global variable
      ‘par.guess’
    Undefined global functions or variables:
      H5Fclose H5Lexists evals h5ls nx par.guess r1 r2

urbach:
for the `r1` and `r2` problem I cannot test a solution right now, because dplyr doesn't really work on my laptop... But I think the solution would be

        random_idcs <- dplyr::filter(random_idcs,                                                                                                                           
                                 random_idcs$r1 != random_idcs$r2) 

?

urbach:
now only this is left

    gevp2cf: possible error in cf_weighted(cf, weight.factor =
      gevp$cf$weight.factor, weight.cosh = gevp$cf$weight.cosh, mass1 =
      gevp$cf$mass1, mass2 = gevp$cf$mass2): unused arguments (mass1 =
      gevp$cf$mass1, mass2 = gevp$cf$mass2)

kostrzewa:
> cyprus_read_loops: no visible global function definition for ‘h5ls’
> cyprus_read_loops: no visible global function definition for
  ‘H5Lexists’
> cyprus_read_loops: no visible global function definition for ‘H5Fclose’

I though I had fixed all of those... weird

kostrzewa:
Yes, I had fixed all of those already.

kostrzewa:
> * checking R code for possible problems ... NOTE
analysis_online: no visible binding for global variable ‘evals’
cyprus_read_loops: no visible global function definition for ‘h5ls’
cyprus_read_loops: no visible global function definition for
  ‘H5Lexists’
cyprus_read_loops: no visible global function definition for ‘H5Fclose’
gevp2cf: possible error in cf_weighted(cf, weight.factor =
  gevp$cf$weight.factor, weight.cosh = gevp$cf$weight.cosh, mass1 =
  gevp$cf$mass1, mass2 = gevp$cf$mass2): unused arguments (mass1 =
  gevp$cf$mass1, mass2 = gevp$cf$mass2)
loop_2pt: no visible binding for global variable ‘r1’
loop_2pt: no visible binding for global variable ‘r2’
set.dfitchi : jacobian: no visible binding for global variable ‘nx’
set.dfitchi : jacobian: no visible binding for global variable
  ‘par.guess’
Undefined global functions or variables:
  H5Fclose H5Lexists evals h5ls nx par.guess r1 r2

Are you sure you weren't looking at an outdated version?

kostrzewa:
I see what happened, I had made the changes in my other branch, falsely assuming that I had cherry-picked them out to `closer-to-cran` at the time...

R/analysis_online.R
```diff
@@ -496,7 +496,7 @@ analysis_online <- function(L, T, t1, t2, beta, kappa, mul,
   }
   if( !missing("evals_id") ) {
     plotcounter <- plotcounter+1
-    ev_pdf_filename <- sprintf("%02d_evals_%02d_%s", plotcounter, evals, filelabel )
+    ev_pdf_filename <- sprintf("%02d_evals_%02d_%s", plotcounter, evals_id, filelabel )
```
kostrzewa: weird... I thought I had fixed this one too...

R/looptools.R
```diff
@@ -155,7 +155,7 @@ loop_2pt <- function(loop_snk, loop_src,
     random_idcs <- expand.grid( 1:dims[1], 1:dims[1], KEEP.OUT.ATTRS = FALSE)
     colnames(random_idcs) <- c("r1", "r2")
     random_idcs <- dplyr::filter(random_idcs,
-                                 r1 != r2)
+                                 random_idcs$r1 != random_idcs$r2)
```
kostrzewa: I think I've found the correct and general solution for this using `rlang::.data`, will push in a few minutes.

kostrzewa: But really, `R CMD check` is outdated in this respect...

kostrzewa: To be honest though, I think I'll just replace this with a `which`

kostrzewa:
Alright, with the addition of `R6` to `Imports` and an explicit `importFrom("R6", "R6Class")`, `R CMD check` passes again. I've also resolved the NOTE due to  `dplyr` lazy evaluation by not using `dplyr` for this simple filtering.

However, in general one has to import `.data` from `rlang` and use `rlang::.data$varname` for the lazy evaluation in a package which should pass `R CMD check` without NOTEs.

kostrzewa:
Note that I've added one somewhat major change in that I've explicitly exported `new_matrixfit` and added documentation for it.

martin-ueding:
I sure what to have `new_matrixfit` exported because it is the only one which can properly use priors, eventually. And we should just keep the name, changing the old name is just not worth it.

kostrzewa:
> I sure what to have new_matrixfit exported because it is the only one which can properly use priors, eventually. And we should just keep the name, changing the old name is just not worth it.

Alright, could you port the documentation, if any, to this branch then? What I added is rather incomplete as it doesn't touch on the new types of fits at all.

urbach:
hmm, something in `correlatormatrix.RData` went wrong when you inserted the imaginary part @kostrzewa 

    Warning: file ‘correlatormatrix.RData’ has magic number 'RDX3'
      Use of save versions prior to 2 is deprecated
    Error in load(zfile, envir = envir) : 
      bad restore file magic number (file may be corrupted) -- no data loaded
    ERROR: lazydata failed for package ‘hadron’

...

kostrzewa:
> hmm, something in correlatormatrix.RData went wrong when you inserted the imaginary part @kostrzewa

since the tests were passing just a few minutes ago, I would guess that this might have to do with a version mismatch of our respective R versions

kostrzewa:
Yeah, the tests are passing here, on QBIG and on travis. I think we should bump the R version requirement to >= 3.5.0 as that's when RDX3 was introduced.

urbach:
> > hmm, something in correlatormatrix.RData went wrong when you inserted the imaginary part @kostrzewa
> 
> since the tests were passing just a few minutes ago, I would guess that this might have to do with a version mismatch of our respective R versions

My

R version 3.4.4

is not soo outdated...

urbach:
> Yeah, the tests are passing here, on QBIG and on travis. I think we should bump the R version requirement to >= 3.5.0 as that's when RDX3 was introduced.

for which actual reason!?

kostrzewa:
> > Yeah, the tests are passing here, on QBIG and on travis. I think we should bump the R version requirement to >= 3.5.0 as that's when RDX3 was introduced.
> 
> for which actual reason!?

In that case we need to generate the sample data elsewhere or investigate if it's possible to write out an older serialisation version with current R versions.

kostrzewa:
> Yeah, the tests are passing here, on QBIG and on travis. I think we should bump the R version requirement to >= 3.5.0 as that's when RDX3 was introduced.

It seems that it's still possible to save in "format version 2" by explicitly specifying the version. I'll replace the example data with data stored in version 2.

martin-ueding:
But R 3.5.0 only came out [2018-04-23](https://stat.ethz.ch/pipermail/r-announce/2018/000628.html), so I'm sure there will be some system where an older version is installed. One can easily write an older version:

```
     saveRDS(object, file = "", ascii = FALSE, version = NULL,
             compress = TRUE, refhook = NULL)
```
```
 version: the workspace format version to use.  ‘NULL’ specifies the
          current default version (3). The only other supported value
          is 2, the default from R 1.4.0 to R 3.5.0.
```

And then we are compatible down to R 1.4.0, which came out [2001-12-19 ](https://en.wikipedia.org/wiki/R_(programming_language)#Milestones). That sounds more realistic.

martin-ueding:
The same `version` argument also works with the `save` function.

kostrzewa:
> But R 3.5.0 only came out 2018-04-23, so I'm sure there will be some system where an older version is installed. One can easily write an older version:

Agreed, 3.5.0 is too recent to make it a requirement. I've replaced the example data instead.

kostrzewa:
@urbach can you confirm that the data now reads fine for you?

urbach:
> @urbach can you confirm that the data now reads fine for you?

yes, I can.

urbach:
so, one thing left, I was in the wrong pull request:

    Undocumented arguments in documentation object 'weight.cf'
      ‘inverse’

    Functions with \usage entries need to have the appropriate \alias
    entries, and all their arguments documented.
    The \usage entries must correspond to syntactically valid R code.
    See chapter ‘Writing R documentation files’ in the ‘Writing R
    Extensions’ manual.

R/new_matrixfit.R
```diff
@@ -238,6 +237,31 @@ NParticleModel <- R6Class(
   )
 )
 
+#' perform a factorising fit of a matrix of correlation functions
```
kostrzewa: @martin-ueding would be great if you could port any additional documentation that you might have from #169

kostrzewa:
why does this contain parts of #186 ?!

R/new_matrixfit.R
martin-ueding: I don't think that there is any new documentation in that branch, I just looked through that.

martin-ueding:
The checking runs through fine on my laptop now (there is a `./check` which does not complain about the git files).

I am not sure about the stuff from #186, did that get merged into `master` and then merged into here perhaps?

martin-ueding:
The commit c9d7c6149aa417d55931f07d64a1b40061040597 is on this branch as well. It seems that somehow @kostrzewa worked on the icf stuff in the `MissingDocu` branch?

Bildschirmfoto_004

Guess we either go with it now or there will be some rebasing fun.

kostrzewa:
yeah, no idea how this came about, certainly not on purpose from my side

on the positive side, it brought the versioning problem with the sample data to light

PR #200, opened 2019-11-14
Bootstrap associative cf arithmetic

kostrzewa:
This is a work-in-progress PR to consistently operate on the bootstrap samples under associative modifications of a correlation function.


kostrzewa:
Note that this contains commits from the `icf_support` branch to avoid duplication of effort. That branch should be merged first via #186

kostrzewa:
There's an annoying issue that we have to take care of somehow. When `tsboot` is used, `cf$cf.tsboot` and `cf$icf.tsboot` contain the original data in their `tseries` member.

On the other hand, when `jackknife.cf` is used, the "hacked-together" `cf.tsboot` does not contain the original data.

In order to work consistently on the samples, I need to address this and see two alternatives:

1) we don't propagate the full result of `tsboot` but only a subset of its members
2) we add the `tseries` member to what `jackknife.cf` attaches

kostrzewa:
I also think that the value of `sim` under `jackknife.cf` should be `fixed`, but that's only cosmetic anyway.

kostrzewa:
Only `symmetrise.cf` left to go.

kostrzewa:
Merge depends on #186

kostrzewa:
ready for review

kostrzewa:
ugh, because I've removed a function, the diff for `R/cf.R` is basically useless to see what has changed... first time I see this here

kostrzewa:
almost a month has passed, requesting a review again

R/cf.R
```diff
@@ -623,15 +671,52 @@ add.cf <- function(cf1, cf2, a = 1.0, b = 1.0) {
...
 
   if( has_icf(cf1) | has_icf(cf2) ){
     stopifnot( has_icf(cf1) & has_icf(cf2) )
+    stopifnot( all(dim(cf1$icf) == dim(cf2$icf) ) )
     cf$icf <- a*cf1$icf + b*cf2$icf
   }
 
-  cf <- invalidate.samples.cf(cf)
+  # now reconstruct the bootstrap samples
+  invalidate.samples.cf(cf)
```
martin-ueding: This statement does not have effect, right?

kostrzewa: The idea, as elsewhere in the code, was to ensure a completely clean slate before re-adding the updated bootstrap samples via the constructor.

martin-ueding: I realize that part. But without the `cf <-` assignment that line is without an effect due to the pass-b-promise nature of R.

kostrzewa: Now I see what you mean, my brain didn't even parse that the assignment had been removed...

Issue #189, opened 2019-10-22
bootstrap.nlsfit only uses correlated chi^2 if a CovMatrix is passed


kostrzewa:
this is despite the fact that samples may be passed

matfischer:
I encountered a weird error using the latest version of `bootstrap.nlsfit`. An error message popped up telling me that boot_r = 0 is not provided. From the documentation I couldn't find any error I made respectively old code which worked back then does not work anymore.

martin-ueding:
@matfischer: The model function needs to accept `...` as a catch-all argument now. That in in the documentation of `bootstrap.nlsfit`: https://github.com/HISKP-LQCD/hadron/blob/master/man/bootstrap.nlsfit.Rd#L13

kostrzewa:
This should also always work now, as long as you fit function accepts `...`, both the `optim` and `minpack.lm` code paths will work now.

PR #205, opened 2019-11-19
Add na.rm for fit routines

j-ostmeyer:
Sometimes it is quite tedious to remove all the NAs in some given data and the bootstrap samples by hand. With the flag `na.rm` in `bootstrap.nlsfit` this problem can be solved automatically. The default remains that NAs lead to breakdown of the fit.


urbach:
that's a useful feature! Can you please also update the documentation accordingly?

Then, the failing checks need to be resolved...

R/bootstrap.nlsfit.R
```diff
@@ -157,16 +159,16 @@ parametric.nlsfit.cov <- function (fn, par.guess, boot.R, y, x, cov, ..., bootst
...
     if(missing(boot.R)) {
       boot.R = 0
     }
-    simple.nlsfit(fn, par.guess, y, x, errormodel, ..., CovMatrix = cov, boot.R = boot.R)
+    simple.nlsfit(fn, par.guess, y, x, errormodel, ..., CovMatrix = cov, boot.R = boot.R, na.rm)
   }
 }
 
-get.errors <- function (useCov, y, dy, dx, CovMatrix, errormodel, bsamples, cov_fn, error) {
+get.errors <- function (useCov, y, dy, dx, CovMatrix, errormodel, bsamples, cov.fn, error) {
```
kostrzewa: why the change of notation for `cov_fn`? This makes it inconsistent with the rest of hadron

R/bootstrap.nlsfit.R
```diff
@@ -639,10 +666,11 @@ bootstrap.nlsfit <- function(fn,
                              use.minpack.lm = TRUE,
                              parallel = FALSE,
                              error = sd,
-                             cov_fn = cov,
+                             cov.fn = cov,
```
kostrzewa: see question above regarding the change of notation here

If anything, `error` should become `err_fn` to increase consistency with the rest of the code.

R/bootstrap.nlsfit.R
martin-ueding: @j-ostmeyer has argued at some point that it should not be inconsistent within the arguments of a single function. But we are actually well past this point with `useCov`, `CovMatrix`, `cov_fn` and `use.minpack.lm`. Modern R (I guess as defined by Hadley Wickham) uses underscores, and that's what I would aim for in the long run.

j-ostmeyer:
Ok, I find it somewhat irritating that some arguments use dots and others use underscores, but I will redo the changes concerning `cov_fn`.

j-ostmeyer:
Solves issue #189 now. Simply pass `CovMatrix = NULL` for covariance treatment without matrix.

j-ostmeyer:
Any further comments? From my point of view this is ready to merge.

kostrzewa:
> Any further comments? From my point of view this is ready to merge.

@martin-ueding should comment on whether the changes in notation for things like `boot_r -> boot.r` affect his analyses

martin-ueding:
I guess it will break stuff in the rho analysis. Also in LuescherAnalysis I use `boot_R` everywhere and then use `boot.R` as a function parameter to hadron functions like `bootstrap.cf`. I can update stuff in the analysis, other people will then have to update their parts eventually.

kostrzewa:
> I guess it will break stuff in the rho analysis. Also in LuescherAnalysis I use boot_R everywhere and then use boot.R as a function parameter to hadron functions like bootstrap.cf. I can update stuff in the analysis, other people will then have to update their parts eventually.

I think this affects only the internal index iterating over "original measurements" and "bootstrap samples" and hence, possibly, the model functions (and I know that in parts of the LuescherAnalysis, `boot_r` is used explicitly to extract a mass from arguments passed via `...` [such as the pion mass in the chiral extrapolation]).

martin-ueding:
Yes, that is the only effect that it has. So it should be easy to fix.

urbach:
> Yes, that is the only effect that it has. So it should be easy to fix.

should we keep for some time both?

martin-ueding:
I don't think that this is possible. We would have to pass `boot.r = boot.r, boot_r = boot.r` to the fit function then. We have instructed user to use `...` to ignore all arguments they do not need, so that would be possible. But that just makes it more ambiguous. So just break it if you want, I can fix my code when it breaks as a consequence. Just make sure that this change is released as a new version such that I can make LuescherAnalysis depend on it. And since it breaks the API it should be major.

kostrzewa:
Agreed, this should bump the hadron version then.

kostrzewa:
> Solves issue #189 now. Simply pass CovMatrix = NULL for covariance treatment without matrix.

so just to confirm the logic 

- `missing(CovMatrix)` -> `useCov = FALSE`
- `CovMatrix == NULL & !missing(bsamples)` -> `useCov = TRUE`

?

j-ostmeyer:
> - missing(CovMatrix) -> useCov = FALSE
> - CovMatrix == NULL & !missing(bsamples) -> useCov = TRUE

Yes.

j-ostmeyer:
> just needs a hadron version bump and we're good to go

Everything is up to date, isn't it?

kostrzewa:
> > just needs a hadron version bump and we're good to go
> 
> Everything is up to date, isn't it?

What I meant is that in the package `DESCRIPTION`, as these changes break some of the interface, the hadron version should be bumped to `2.5.0`.

martin-ueding:
If we break the API in minor versions I cannot ridicule the versioning policy of the GNOME project any more :laughing:.

kostrzewa:
Alright, push it up to 3 then... We'll talk again in three months when we've reached version 361.1.0 :)

martin-ueding:
I don't care that much, we can just leave it at that. As long as it is monotonically increasing I can update LuescherAnalysis. But in general I think having to bump the major version is a good deterrent to break the API.

kostrzewa:
> But in general I think having to bump the major version is a good deterrent to break the API.

I agree!

kostrzewa:
ps: to annoy you, I'll make some random changes in the subminor version at some random points in time in the future :)

martin-ueding:
That's fine, it's an improvement over random changes with no version number bump whatsoever. And I'd love to see some Flash player version numbers, say 32.0.0.293. Or RStudio has 1.2.0.5001. That's the level of detail that we should have. And we should support multiple different minor versions for decades. But that's nothing against the version numbering scheme of GNOME, they just rid themselves with all semantic version numbers and do their own thing. We shouldn't let ourselves be limited by the sole existence of [semver](https://semver.org/).

martin-ueding:
> @martin-ueding should comment on whether the changes in notation for things like boot_r -> boot.r affect his analyses

As expected this has broken the analysis and the fallout for this was that @matfischer lost around a nice half day of work. :+1:

PR #209, opened 2019-12-14
Show all points in plot for new_matrixfit

### LINKED ISSUE #208: new_matrixfit does not show all points in plot
By construction the `new_matrixfit` only passes the time slices of the correlator that are actually fitted to `bootstrap.nlsfit`. Then in consequence `plot.bootstrapfit` will only plot the points that have been used, which is not that useful. Rather we should have all the data in the plot. This will require changes to `new_matrixfit`, `bootstrap.nlsfit` and `plot.bootstrapfit`.

martin-ueding:
Fixes #208.


martin-ueding:
Here you can see that it now produces the same result as `plot.matrixfit` did:

[new_matrixfit_all_points.pdf](https://github.com/HISKP-LQCD/hadron/files/3963724/new_matrixfit_all_points.pdf)

martin-ueding:
There is an additional complication which just surfaced. The correlator fit models that we have for the `new_matrixfit` functions are just defined on the time slices that are actually fitted on. This arises from the use of `sign_vec` and `ov_sign_vec`. These vectors contain the overall sign and back-propagating signs for all time slices for all correlators in the matrix. Their length for T time slices and N correlators is exactly T * N. They are needed because different parts of the matrix might have different overall or relative signs in them.

Whenever one tries to evaluate the model on a different number of supports, R will warn about non-matching lengths of the vectors. This comes in the multiplication `self$ov_sign_vec * par[self$parind[, 1]]` and `self$sign_vec * exp(-par[1] * (self$time_extent - x))`. The `length(x)` must match `sign_vec` and `ov_sign_vec`. This is not the case in `plot.bootstrapfit` and only did not raise any issues because the number of supports has been chosen *by coincidence* to be a divisor of 11. Changing the number of supports will make it spew a lot of warnings, which become critical when the values are not just all `+1`.

I deem the model ill defined. The `new_matrixfit` function takes the user supplied `sym.vec` and `neg.vec` and inflates them with the given number of time slices to fit. This action is done too early, it must be done in the model such that it is matched to `length(x)` at evaluation time. Only this way we can properly evaluate the model on all supports.

In the old `plot.matrixfit` function this issue was solved by having yet another definition of the fit model and fiddling with the signs there manually and evaluating the model for each correlator separately.

Going forward I want to put the generation of the `sign_vec` and `ov_sign_vec` from `sym.vec` and `neg_vec` into the models.

martin-ueding:
I've made these changes now, and they seem to work. Of course this makes `new_matrixfit` even slower because there is additional indirection and the redundant computation of these vectors and the `parind` matrix. Perhaps one could add some mechanism to cache them.

martin-ueding:
The error computation between the old and new ratio plot seems to be different. I do not understand why that is the case yet.

Bildschirmfoto_012

Bildschirmfoto_013

R/bootstrap.nlsfit.R
```diff
@@ -1044,3 +1098,43 @@ plot.bootstrapfit <- function(x, ..., col.line="black", col.band="gray", opacity
   ## plot the fitted curve on top
   lines(x=X, y=Y, col=col.line, lty=lty, lwd=lwd)
 }
+
+residual_plot <- function (x, ...) {
+  UseMethod("residual_plot", x)
```
j-ostmeyer:
```suggestion
  UseMethod("residual_plot", x, ...)
```

martin-ueding: It works without that, I believe that `UseMethod` magically forwards all the arguments.

kostrzewa:
> The error computation between the old and new ratio plot seems to be different. I do not understand why that is the case yet.

I think you might have forgotten to divide the error by the fit function and have divided just the central values.

kostrzewa:
That of course assumes a lack of correlation between the correlator and the fit, which is rather silly. If you computed the errors of the ratio via the bootstrap, then your plot is correct and the original one is simply much more wrong that we thought.

kostrzewa:
On the other hand, I think the original thinking was that if the correlator, with its statistical error scaled simply by the central value of the fit function, overlaps the fit result within errors, that's a good fit, whereas if one computes the correlated ratio, one will basically never have the indication of a reasonable fit because in the signal region, the correlator and the fit function should be extremely correlated, up to fluctuations from time slice to time slice.

R/bootstrap.nlsfit.R
j-ostmeyer: I don't see how this is possible and would add the dots in order to avoid this confusion in future. But do as you like.

kostrzewa: `UseMethod` only takes two arguments, the dots will be ignored anyway. Under the hood, magic happens and a function call is generated which matches the signature of the generic, which has `x` as its first argument and `...` as its second. It's really weird... see *Technical Details* in `?UseMethod`

R/bootstrap.nlsfit.R
```diff
@@ -1044,3 +1098,43 @@ plot.bootstrapfit <- function(x, ..., col.line="black", col.band="gray", opacity
...
+  }
+  do.call(plotwitherror, plot_args)
+  
+  plot_args <- list(x=x$x[x$mask], y=residual_val[x$mask], dy=residual_err[x$mask], ..., rep = TRUE)
+  if(x$errormodel == "xyerrors") {
+    plot_args$dx <- x$dx[x$mask]
+  }
+  do.call(plotwitherror, plot_args)
+  
+  abline(h = 1, col = 'gray70')
```
j-ostmeyer: We should at least add the possibility to draw an errorband.

kostrzewa: An error band of what?

j-ostmeyer: Of the function. In general this would not even be rectangular, so it could be of interest.

kostrzewa: But we're not showing the function here, it's the data divided by the fit function. How do you define this in between the points?

martin-ueding:
According to our discussion I have now just divided by the model prediction value instead of the bootstrap samples.

```diff
--- a/R/bootstrap.nlsfit.R
+++ b/R/bootstrap.nlsfit.R
@@ -1120,7 +1120,7 @@ residual_plot.bootstrapfit <- function (x, ..., error_fn = sd, operation = `/`)
   prediction_boot <- do.call(rbind, lapply(1:nrow(x$t), prediction_boot_fn))
   
   residual_val <- operation(x$y, prediction_val)
-  residual_boot <- operation(x$bsamples[, 1:length(x$y)], prediction_boot)
+  residual_boot <- operation(x$bsamples[, 1:length(x$y)], prediction_val)
   str(residual_boot)
   residual_err <- apply(residual_boot, 2, error_fn)
```

But the errors in the new plot look super strange:

Bildschirmfoto_014

Bildschirmfoto_015

Bildschirmfoto_016

That looks even worse than before. I have to understand what the old code actually did exactly.

kostrzewa:
You need to divide the errors too.

kostrzewa:
Are you sure that the division is carried out correctly on the bootstrap samples? matrix divided by vector ...

martin-ueding:
I don't divide the errors, I divide the bootstrap samples and then take the standard deviation. And yes, the dimensions work out. Adding a `t()` just to check breaks it.

martin-ueding:
Ah, one just has to put the `t()` somewhere else. It is very sad that 400 is a multiple of 25. I'd greatly appreciate if R wasn't trying to be so clumsy while pretending to be helpful 🤬

kostrzewa:
My expectation would be that R divides the columns, repeating `prediction_val` to match the number of rows of the bootstrap samples. Could you confirm with apply?

kostrzewa:
> Ah, one just has to put the `t()` somewhere else. It is very sad that 400 is a multiple of 25. I'd greatly appreciate if R wasn't trying to be so clumsy while pretending to be helpful 🤬

Yeah, I try to avoid making use of vector repetition if at all possible. It leads to unclear code. If I do rely on it, I explicitly check that the thing I'm repeating has length 1.

martin-ueding:
Nah, I've added enough `t()` such that it works. The next person should have fun with it as well. 🥳

The errors now work as expected. I have also added the error band which is `prediction_boot / prediction_val`.

Bildschirmfoto_017

So how do we actually read this?

- Comparing data with the unity line we see that the data is compatible with the central mass value extracted.
- Comparing the error band with the central values of the data points tells us that the data would be well within the prediction uncertainty of the model, the _p_-value of the fit should be rather high.
- Comparing the error band with the errors of the data seems tempting. Does that tell us that all the uncertainty of the data has been properly reflected in the uncertainty of the model parameters? What would it mean if one is much larger than the other?

martin-ueding:
I'm happy with it now, we can merge it.

j-ostmeyer:
> Comparing the error band with the errors of the data seems tempting. Does that tell us that all the uncertainty of the data has been properly reflected in the uncertainty of the model parameters? What would it mean if one is much larger than the other?

An errorband as large as the individual errors tells us that the points are extremely correlated (what a surprise). Basically anyone of them carries the complete information. An errorband smaller than the individual errors would mean less correlation up to the optimum of `individual error / sqrt(N)`. A larger error band usually indicates that some fits did not converge. This should not happen.

kostrzewa:
The tests currently fail so that needs to be checked before this can be merged.

There's one more display behaviour that we don't really test and that is broken by the way `bootstrap.nlsfit` plots the fit result: `bootstrap.nlsfit` can be used without problems to perform global fits to multiple data defined at the same 'x' and described by a common model (sharing at least some parameters) because the fit function is not restricted in any way and can deal with the parameters and passed x values as it pleases. This includes, for example, a matrix of correlators with different interpolating operators for the same quantum numbers which share the energy but have different amplitudes.

I guess this was not really foreseen in the design of `bootstrap.nlsfit`, although it is a very common usecase. 

In `plot.bootstrapfit` one can (almost) cause the correct behaviour by using `plot.range` to restrict plotting to a certain x region. However, instead of passing `x` directly, what should be passed is a mask of indices for the purpose of plotting, such that a particular subset of `x` values can be selected corresponding to the part of the model that one would want to plot. This is a hard problem to solve in general because one will also need to select a subset of parameters to produce the appropriate prediction... currently I don't have an easy fix for this.

As it is right now, if `plot.boostrapfit` is used to plot a factorising fit to a correlator matrix, the results will be nonsensical.

A possible workaround would be to offset the x values for the different pieces of the data sharing the same x values and to undo this offset in the fit function. That way, the fit function would describe the data correctly, but from the point of view of `bootstrap.nlsfit`, all x values would be unique...

R/bootstrap.nlsfit.R
```diff
@@ -1044,3 +1098,52 @@ plot.bootstrapfit <- function(x, ..., col.line="black", col.band="gray", opacity
...
+  
+  # The same is done for the bootstrap samples
+  prediction_boot_fn <- function (boot.r) {
+    par <- x$t[boot.r, 1:npar]
+    do.call(x$fn, c(list(par = par, x = x$x, boot.r = boot.r), x$tofn))
+  }
+  prediction_boot <- do.call(rbind, lapply(1:nrow(x$t), prediction_boot_fn))
+  
+  residual_val <- operation(x$y, prediction_val)
+  residual_boot <- t(operation(t(x$bsamples[, 1:length(x$y)]), prediction_val))
```
kostrzewa: could you put a comment here to explain the double transpose?

j-ostmeyer:
> However, instead of passing x directly, what should be passed is a mask of indices for the purpose of plotting

I strongly disagree. One could of course add an additional mask, but I need the region for extrapolations of the fit function beyond the fitted values.

kostrzewa:
> A possible workaround would be to offset the x values for the different pieces of the data sharing the same x values and to undo this offset in the fit function. That way, the fit function would describe the data correctly, but from the point of view of bootstrap.nlsfit, all x values would be unique...

Another workaround (which I'm using right now in another project) is to make the fit function smarter, passing flags via `...` which dictate which part of the global model is to be evaluated for the given set of passed 'x' values.

kostrzewa:
> > However, instead of passing x directly, what should be passed is a mask of indices for the purpose of plotting
> 
> I strongly disagree. One could of course add an additional mask, but I need the region for extrapolations of the fit function beyond the fitted values.

Agreed, that's useful functionality.

kostrzewa:
> > A possible workaround would be to offset the x values for the different pieces of the data sharing the same x values and to undo this offset in the fit function. That way, the fit function would describe the data correctly, but from the point of view of bootstrap.nlsfit, all x values would be unique...
> 
> Another workaround (which I'm using right now in another project) is to make the fit function smarter, passing flags via `...` which dictate which part of the global model is to be evaluated for the given set of passed 'x' values.

But this is further complicated by the fact that the `...` of `plot.bootstrapfit` if of course not passed to the fit function but `plotwitherror`...

R/bootstrap.nlsfit.R
```diff
@@ -1120,10 +1120,41 @@ residual_plot.bootstrapfit <- function (x, ..., error_fn = sd, operation = `/`)
   prediction_boot <- do.call(rbind, lapply(1:nrow(x$t), prediction_boot_fn))
   
   residual_val <- operation(x$y, prediction_val)
+  # We want to subtract or divide (depending on the given `operation`) the
```
kostrzewa: Very funny... One line would have been enough...

martin-ueding: I've made it a one-liner now.

kostrzewa: Shame, I liked the outburst. Probably better this way though...

martin-ueding:
> But this is further complicated by the fact that the `...` of `plot.bootstrapfit` if of course not passed to the fit function but `plotwitherror`...

One could refrain from using `...` in the `plot.bootstrapfit` and instead have `extra_plot_args = list(), extra_model_args = list()`. Internally one would use `do.call` do pass those additional arguments. But that would break the S3 API of `plot`, which needs to have `plot(x, y, ...)`. But we could just do `extra_model_args` and have `...` for all those graphical parameters. Pushing them into a list would also break existing code.

kostrzewa:
> > But this is further complicated by the fact that the `...` of `plot.bootstrapfit` if of course not passed to the fit function but `plotwitherror`...
> 
> One could refrain from using `...` in the `plot.bootstrapfit` and instead have `extra_plot_args = list(), extra_model_args = list()`. Internally one would use `do.call` do pass those additional arguments. But that would break the S3 API of `plot`, which needs to have `plot(x, y, ...)`. But we could just do `extra_model_args` and have `...` for all those graphical parameters. Pushing them into a list would also break existing code.

I would say that for now we live with it and when we come to the situation where we actually need the factorising fits, we implement a `plot.new_matrixfit` which does the right thing. No need to invest time now.

kostrzewa:
Shall I press the button or is there something else that needs to be done?

R/bootstrap.nlsfit.R
martin-ueding: Oh, I read your comment as ironic. I've reverted the reverting, so we have it in the code now.

Issue #210, opened 2019-12-16
bootstrap.nlsfit, get.errors, correlated fit, CovMatrix = NULL ==> incorrect dy


kostrzewa:
https://github.com/HISKP-LQCD/hadron/blob/6867cc2f5ee45a3ed958e54aded50543ba4a3ae9/R/bootstrap.nlsfit.R#L195

When a correlated fit is done and `CovMatrix` is passed as `NULL`,  `W` is computed from the bootstrap samples and the input data errors are in turn computed via

```r
dydx <- 1.0/diag(W)
```

~This all seems fine, except that the errors that are stored as part of the return value now depend on the quality of the inverse of the covariance matrix. When the inversion is problematic, values of `dy` stored in the return value do not correspond to the actual statistical errors of the data...~

This is just simply wrong...

kostrzewa:
Besides, this is not correct in the first place. The errors, if they are taken from the covariance matrix, should be taken from the covariance matrix.

kostrzewa:
@NikSchlage It seems that you made this particular choice, what was the reasoning there?

kostrzewa:
@NikSchlage did you use this in production for published / to be published plots?

NikSchlage:
@kostrzewa I did not make that choice, I only renamed dy -> W in line 195.

NikSchlage:
@kostrzewa I also think that the inverse cov matrix is wrong. And yes, I used this already existing code (with my prior modifications) for my plot production.

kostrzewa:
@NikSchlage Indeed, looking back in history this was introduced multiple times in different guises and moved around in the code.

Note that it might be a little unclear whether this was or was not present in versions of hadron used for production (unless you have the exact version of the code that you used lying around).

I think this might have found it's way into the code around July 2018 or so, at which point it was correct (as it dealt with a particular edge case in the logic tree). At some point beyond that, it is very difficult to figure out at which point, it moved out of this edge case. Beyond that it's even more difficult to follow what happened to the logic tree.

I would say there might be a possibility that the plots that you produced in fact had correct errors, but it's a bit hard to say.

PR #211, opened 2019-12-16
bootstrap.nlsfit: compute dydx from the diagonal of CovMatrix rather …

kostrzewa:
…than from the diagonal of its inverse

### DIFF
--- a/R/bootstrap.nlsfit.R
+++ b/R/bootstrap.nlsfit.R
@@ -640,8 +640,13 @@ simple.nlsfit <- function(fn,
 #' @param CovMatrix complete variance-covariance matrix of dimensions
 #' \code{c(length(y), length(y))} or \code{c(length(y)+length(x),
 #' length(y)+length(x))} depending on the errormodel. Pass `NULL` if the matrix
-#' has to be calculated from the `bsamples`. If missing, uncorrelated fit will
-#' be used.
+#' has to be calculated from the `bsamples`. In that case, if the number of
+#' boostrap samples is small compared to the number of variables, singular value
+#' decomposition with small eigenvalue replacement will be used (see \link{invertCovMatrix})
+#' to attempt a clean inversion.
+#' In case a variance-covariance matrix is passed, the inversion will simply be attempted
+#' using \code{solve} on the Cholesky decomposition.
+#' Finally, if `CovMatrix` is missing, an uncorrelated fit will be performed.
 #' @param mask logical or integer index vector. The mask is applied to select the observations from the data that are to be used in the fit. It is applied to `x`, `y`, `dx`, `dy`, `bsamples` and `CovMatrix` as applicable.
 #' @param use.minpack.lm use the \code{minpack.lm} library if available. This
 #' is usually faster than the default \code{optim} but somtimes also less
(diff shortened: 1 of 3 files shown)


kostrzewa:
This makes more sense than #210, right?

R/bootstrap.nlsfit.R
```diff
@@ -182,6 +182,7 @@ get.errors <- function (useCov, y, dy, dx, CovMatrix, errormodel, bsamples, cov_
     }
 
     if (is.null(CovMatrix)) {
+      CovMatrix <- cov_fn(bsamples)
       InvCovMatrix <- try(invertCovMatrix(bsamples, boot.l = 1, boot.samples = TRUE, cov_fn = cov_fn), silent = TRUE)
```
j-ostmeyer:
```suggestion
      InvCovMatrix <- try(solve(CovMatrix), silent = TRUE)
```
Or even better calculate the covariance matrix, if missing. Then invert and cholesky decompose. No redundancies in the `if...else`.

j-ostmeyer:
Apart from the  upper remark I totally agree.

R/bootstrap.nlsfit.R
kostrzewa: The reason that I didn't unify the two logic branches was the following: I assume that when one bothers to pass a hand-crafted covariance matrix, one might have made sure that it can be inverted naively. On the other hand, when the covariance matrix is computed simply from the bootstrap samples, one might want to use the SV decomposition with small eigenvalue replacement. What do you think?

kostrzewa: I've added some comments and extended the documentation somewhat to clarify things.

martin-ueding:
This function has reached the point again where it violates SRP and ISP from [SOLID](https://en.wikipedia.org/wiki/SOLID). Basically there are multiple distinct use cases for this function:

1. Having just the bootstrap samples and doing an uncorrelated
1. or correlated fit.
1. Having the errors and doing a parametric bootstrap first. Then we have both `dy` and the samples and want to pass both because the error estimate from the bootstrap samples is worse than what we had before.
1. When we have the covariance matrix and do parametric bootstrap we also do not want to have it estimated again.
1. We have the bootstrap samples and computed the covariance matrix, but crafted it in such a way that it is easier to invert within the function.

And then all these things with priors. And all these things with uncertainty on the independent variable. Basically that's like 20 different variants and there is just one interface, so that's against ISP. And I doubt that SRP is anywhere close here.

kostrzewa:
> And then all these things with priors. And all these things with uncertainty on the independent variable. Basically that's like 20 different variants and there is just one interface, so that's against ISP. And I doubt that SRP is anywhere close here.

We could relatively easily split the (user) interface into several driver functions. To some extent, that is already present via the `simple.nlsfit`, `parametric.nlsfit` and `parametric.nlsfit.cov` shortcuts. The resulting driver functions would contain all the logic to set up the fit for their particular variant. This would necessarily entail some duplication, but it would remove a lot of branching elsewhere.

In the end, I think, this boils down to removing `dy` and `dx` from `bootstrap.nlsfit`. (as one should use `simple.nlsfit` if one has errors but no samples).

Then, `bootstrap.nlsfit` narrows in scope: fits with bootstrap samples either

1) uncorrelated
2) automatic CovMatrix
3) hand-crafted covariance matrix

Any number of these driver functions can be written. For example, when a mask is passed, that could be one driver function. Similarly, when priors are used, that may be another driver function.

R/bootstrap.nlsfit.R
j-ostmeyer: I see. It was not clear to me that the algorithms differ that significantly.

PR #212, opened 2019-12-17
Bootstrap nlsfit constraints

kostrzewa:
This adds support for `upper` and `lower` constraints to be passed to the minimisers. There might be some subtleties when these things interact with priors...

### DIFF
--- a/R/bootstrap.nlsfit.R
+++ b/R/bootstrap.nlsfit.R
@@ -874,6 +874,13 @@ bootstrap.nlsfit <- function(fn,
     bsamples <- rbind(Yp, bsamples)
   }
   
+  if (length(upper) < length(par.Guess)) {
+    upper <- c(upper, rep(+Inf, times = length(par.Guess) - length(upper)))
+  }
+  if (length(lower) < length(par.Guess)) {
+    lower <- c(lower, rep(-Inf, times = length(par.Guess) - length(lower)))
+  }
+  
   wrapper <- set.wrapper(fn, gr, dfn, par.guess, errormodel, useCov, W, x, ipx, lm.avail, maxiter, success.infos, na.rm, priors, priors.avail, lower, upper)
   
   ## now the actual fit is performed
--- a/R/bootstrap.nlsfit.R
+++ b/R/bootstrap.nlsfit.R
@@ -104,10 +104,19 @@ parametric.bootstrap.cov <- function (boot.R, x, cov, seed) {
 #' fit.result <- parametric.nlsfit(fn, c(1, 1), boot.R, value, dvalue, x, dx)
 #' summary(fit.result)
 parametric.nlsfit <- function (fn, par.guess, boot.R, y, dy, x, dx,
+                               lower, upper,
                                ..., bootstrap=TRUE, na.rm = FALSE) {
   stopifnot(length(x) == length(y))
   stopifnot(missing(dx) || length(dx) == length(x))
   stopifnot(missing(dy) || length(dy) == length(y))
+  if( missing(lower) ){
+    lower <- rep(x = -Inf, times = length(par.guess))
+  }
+  stopifnot(length(lower) == length(par.guess))
+  if( missing(upper) ){
+    upper <- rep(x = +Inf, times = length(par.guess))
+  }
+  stopifnot(length(upper) == length(par.guess))
 
   if (missing(dx)) {
     values <- y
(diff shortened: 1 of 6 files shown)


kostrzewa:
@NikSchlage can you predict what might happen when priors are used?

kostrzewa:
I don't see why the example fails even though tests go through fine...

R/bootstrap.nlsfit.R
```diff
@@ -104,10 +104,19 @@ parametric.bootstrap.cov <- function (boot.R, x, cov, seed) {
...
 #' summary(fit.result)
 parametric.nlsfit <- function (fn, par.guess, boot.R, y, dy, x, dx,
+                               lower, upper,
                                ..., bootstrap=TRUE, na.rm = FALSE) {
   stopifnot(length(x) == length(y))
   stopifnot(missing(dx) || length(dx) == length(x))
   stopifnot(missing(dy) || length(dy) == length(y))
+  if( missing(lower) ){
+    lower <- rep(x = -Inf, times = length(par.guess))
+  }
```
martin-ueding: Why don't you just give this as a default value in the function signature?

kostrzewa: why easy and clean when hard is an option? I was just not thinking about the fact that  the default arguments can be rather elaborate in R...

martin-ueding:
Here is your error from Travis CI:

```
> ### Name: bootstrap.nlsfit
> ### Title: Bootstrap a non-linear least-squares fit
> ### Aliases: bootstrap.nlsfit
> 
> ### ** Examples
> 
> ## Declare some data.
> value <- c(0.1, 0.2, 0.31)
> dvalue <- c(0.01, 0.01, 0.015)
> x <- c(1, 2, 3)
> dx <- c(0.1, 0.1, 0.1)
> boot.R <- 1500
> 
> fn <- function (par, x, boot.r, ...) par[1] + par[2] * x
> 
> ## Before we can use the fit with this data, we need to create bootstrap
> ## samples. We do not want to use the correlation matrix here. Note that you
> ## can simply use the parametric.nlsfit function as a convenient wrapper of
> ## the two steps.l
> bsamples <- parametric.bootstrap(boot.R, c(value, x), c(dvalue, dx))
> head(bsamples)
           [,1]      [,2]      [,3]      [,4]     [,5]     [,6]
[1,] 0.09373546 0.2085004 0.3210867 1.1557954 1.938117 2.920520
[2,] 0.10183643 0.1907469 0.3157991 0.9270703 1.889058 3.060608
[3,] 0.09164371 0.2089358 0.3294460 0.8496049 1.782966 2.893753
[4,] 0.11595281 0.1905899 0.2979466 0.9433213 1.996870 3.101920
[5,] 0.10329508 0.2053895 0.2859606 0.7895546 1.973960 3.017761
[6,] 0.09179532 0.1981803 0.3239988 1.0530732 2.053443 2.896903
> 
> fit.result <- bootstrap.nlsfit(fn, c(1, 1), value, x, bsamples)
Error in minpack.lm::nls.lm(par = par, fn = fitchi, y = y, jac = dfitchi,  : 
  length(lower) must be equal to length(par)
Calls: bootstrap.nlsfit -> wrapper
Execution halted
```

The problem is that this examples uses errors on the independent variable and therefore there are more fit parameters than `par.guess` has. It needs to be the length of `par.Guess` because that also contains the starting values for the independent variable. A similar thing will happen with the priors. I will just add a clause that extends both with ±Inf as needed because we do not want to constrain the stuff that has constrains by priors already.

martin-ueding:
That's fixed, but it still occurs in the simple fit:

```
> ### Name: simple.nlsfit
> ### Title: NLS fit with without bootstrap
> ### Aliases: simple.nlsfit
> 
> ### ** Examples
> 
> ## Declare some data.
> value <- c(0.1, 0.2, 0.3)
> dvalue <- c(0.01, 0.01, 0.015)
> x <- c(1, 2, 3)
> dx <- c(0.1, 0.1, 0.1)
> 
> fn <- function (par, x, ...) par[1] + par[2] * x
> 
> fit.result <- simple.nlsfit(fn, c(1, 1), value, x, "xyerrors", dy=dvalue, dx=dx)
Error in minpack.lm::nls.lm(par = par, fn = fitchi, y = y, jac = dfitchi,  : 
  length(lower) must be equal to length(par)
Calls: simple.nlsfit -> wrapper
Execution halted
```

PR #213, opened 2019-12-17
Usable n_particles model

martin-ueding:
I am not aware of an existing possibility to use the `n_particles` model with real data. Perhaps it is somewhere buried in the monolithic Rmarkdown document, I don't know.

`new_matrixfit` has a new parameter which accepts the energy levels of higher states and feeds these as priors into `bootstrap.nlsfit`. In a test with a constant pollution it seems to yield results that don't look too wrong.

[multi_particle_fit_test.pdf](https://github.com/HISKP-LQCD/hadron/files/3974082/multi_particle_fit_test.pdf)

I will try it with the actual 3pi data and see how that goes.


martin-ueding:
Somewhere during the night I realized that I fixed the wrong parameter. Now this is fixed and the model explains the correlator perfectly.

[multi_particle_fit_test.pdf](https://github.com/HISKP-LQCD/hadron/files/3978382/multi_particle_fit_test.pdf)

martin-ueding:
There is still an dimensionality issue with the covariance matrix, I need to fix the correlated fit.

martin-ueding:
The priors must be incorporated into the covariance matrix. But the mask must not cut these away. So I just added another kludge and skipped documenting it in the code. Basically if one wants to have a correlated fit with priors one has to pass `CovMatrix` for it to enable a correlated fit. But then it uses `cov_fn(cbind(bsamples[mask], priors$boot))` to compute the new one.

martin-ueding:
I will also merge this as it seems to work and nobody complained so far.

PR #215, opened 2019-12-19
Prony or Hankel method to extract multiple energy levels from a single correlator

urbach:
this is the Prony method which I have tried to provide now in a usable version.


R/hankel.R
```diff
@@ -102,65 +111,139 @@ gevp.hankel <- function(cf, t0=1, deltat=1, n, N, eps=0.0001, range=c(0,1),
...
-bootstrap.hankel <- function(cf, t0, deltat=1, n, N, id=c(1), range=c(0,1), eps=0.001,
-                             element.order=c(1,2,3,4), submatrix.size=1) {
-  if(t0+2*n+deltat >= N) {
-    stop("t0+2*n+1+deltat must be smaller than N\n")
-  }
+bootstrap.hankel <- function(cf, t0, n=2, N, id=c(1), range=c(0,1), eps=0.001) {
+
+  stopifnot(inherits(cf, 'cf_meta'))
+  stopifnot(inherits(cf, 'cf_boot'))
+  if(missing(N)) N <- cf$Time/2+1
```
kostrzewa: I guess this should fail if the `cf` is not symmetrised, right?

urbach:
> 
> 
> 
> >  #' @family hankel
> -bootstrap.hankel <- function(cf, t0, deltat=1, n, N, id=c(1), range=c(0,1), eps=0.001,
> -                             element.order=c(1,2,3,4), submatrix.size=1) {
> -  if(t0+2*n+deltat >= N) {
> -    stop("t0+2*n+1+deltat must be smaller than N\n")
> -  }
> +bootstrap.hankel <- function(cf, t0, n=2, N, id=c(1), range=c(0,1), eps=0.001) {
> +
> +  stopifnot(inherits(cf, 'cf_meta'))
> +  stopifnot(inherits(cf, 'cf_boot'))
> +  if(missing(N)) N <- cf$Time/2+1 
> 
> I guess this should fail if the `cf` is not symmetrised, right?

not neccessarily, it would still work, wouldn't it?

j-ostmeyer:
The method is not applicable at all for `cosh`-type correlators, i.e. back-propagating parts destroy it. Correct?

urbach:
Yes,  I think this is correct. But the idea is, of course, that this works in regions of time, where the back propagating part is not important.
Carsten Urbach, www.carsten-urbach.eu

martin-ueding:
There is a build error and I cannot fix it in your repository, so I will open a new pull request.

PR #217, opened 2019-12-19
Add mask to simple.nlsfit

j-ostmeyer:
... and simplify code a bit.

### DIFF
--- a/R/bootstrap.nlsfit.R
+++ b/R/bootstrap.nlsfit.R
@@ -477,29 +477,29 @@ simple.nlsfit <- function(fn,
 
   useCov <- !missing(CovMatrix)
 
+  if (use.minpack.lm) {
+    lm.avail <- requireNamespace('minpack.lm')
+  } else {
+    lm.avail <- FALSE
+  }
+
+  all.errors <- get.errors.wo.bootstrap(useCov, y, dy, dx, CovMatrix, errormodel)
+
   ## Apply the mask as in bootstrap.nlsfit.
   if (!missing(mask)) {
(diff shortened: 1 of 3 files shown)


martin-ueding:
Your pull request comes from your own fork. This means that I cannot push additional commits before merging it. It would be nicer if you just pushed this into a branch in this repository.

martin-ueding:
I have just tried this for my fits with yerrors, some with priors, some without. And they now fail with this (in case of priors):

```
 $ message: chr "the leading minor of order 3 is not positive definite"
 $ call   : language chol.default(CovMatrix)
```

Or that one (not using priors)

```
 $ message: chr "Variance-covariance matrix could not be inverted!"
 $ call   : language inversion.worked(InvCovMatrix)
```

Some fail with the current `master` version, but that is okay since it is due to fit ranges. With your version here it seems that all the fits fail. So there has been some change which breaks `bootstrap.nlsfit` for these cases.

We should add more examples to secure this.

j-ostmeyer:
> $ message: chr "Variance-covariance matrix could not be inverted!"
> $ call   : language inversion.worked(InvCovMatrix)

This looks as if your original Cov-matrix is too badly conditioned and only the one with applied mask can be inverted. So I will move the error calculation back after the mask application.

j-ostmeyer:
Note that the way it is implemented now we avoid redundant error calculations.

j-ostmeyer:
> Your pull request comes from your own fork. This means that I cannot push additional commits before merging it. It would be nicer if you just pushed this into a branch in this repository.

Just did so. See #221 .

Issue #219, opened 2019-12-23
chisqr in `matrixfit` takes only integer values!? 


urbach:
Try

    example(matrixfit)
    fitres$t[,4]

or look at the QQ-Plot. That is odd and certainly different from before...!

kostrzewa:
The problem is that the last element of `$t` is not `chisq` per bootstrap sample but the number of iterations that the minimizer took. (see `fit.formatrixboot`)

There'll be a PR in a few seconds to fix the qqplot.

kostrzewa:
As a side-node, there is also a bug in the function which, I think, only didn't surface because of R's partial matching. I'll fix that one too.

kostrzewa:
One more reason to always prefer data frames over matrices for return values...

PR #220, opened 2019-12-23
plot.matrixfit: bootstrap samples of chisq are now in second-to-last …

kostrzewa:
…column of  matrixfit member


kostrzewa:
fixes #219

PR #221, opened 2019-12-23
Mask to simple fit

j-ostmeyer:
... and simplify code a bit.


j-ostmeyer:
See conversation in #217 .

j-ostmeyer:
What is the status of priors in simple.nlsfit? One should either remove the stop-statement (if implemented), or the code segments with priors (if not), or implement it.

R/removeTemporal.cf.R
```diff
@@ -378,10 +380,10 @@ weight.cf <- function (cf, energy_difference_val, energy_difference_boot,
 #' @inheritParams weight.cf
 weight_shift_reweight.cf <- function (cf, energy_difference_val, energy_difference_boot, cosh_factor) {
   cf <- weight.cf(cf, energy_difference_val, energy_difference_boot,
-                  cosh_factor, 0, FALSE)
+                  cosh_factor, 0, TRUE)
```
kostrzewa: is it normal that this PR touches this?

kostrzewa: ah, this is outdated

j-ostmeyer:
This breaks down somewhere in Hankel. Is the problem solved by now and simply not merged into this branch, or it is something more complicated?

urbach:
no, this is still the `multi_particle_fit_test.Rmd` vignette, not hankel.

urbach:
this is a crude fix, by not evaluating the problematic chunks. This needs to be remembered, of course, for later...

urbach:
now package `stringi` is the problem, like for hankel... Something must have changed...

urbach:
> This breaks down somewhere in Hankel. Is the problem solved by now and simply not merged into this branch, or it is something more complicated?

its solved in the hankel branch, though not realted to the hankel
method. For some reason, bibtex does not work anymore, which I also do
not understand at all. Still worked for me on Monday...

j-ostmeyer:
No idea, what happens there. Could you be a bit more specific, please? My fits work.

j-ostmeyer:
I hope that everything is fine now. Just so you know: The code never gave correct results for xy-errors with mask or priors. It simply did not crash out of courtesy of R. Vectors of different lengths can be multiplied...

martin-ueding:
Oh wow, good that you found these issues! Luckily I only use the combinations (y-errors, covariance, priors, mask) and (xy-errors, covariance) but not xy-errors with either mask or covariance.

martin-ueding:
I have just tested this on my 3pi Lüscher Analysis and it produces the exact same results.

Issue #226, opened 2020-01-05
uwerr.cf is broken


martin-ueding:
Somehow the `uwerr.cf` is broken on `master`:

```
> library(hadron)

> uwerr.cf(samplecf)
Error in FUN(newX[, i], ...) : object 'uw_temp' not found

> traceback()
6: FUN(newX[, i], ...)
5: apply(X = cf$cf, MARGIN = 2L, FUN = uw_wrapper)
4: t(apply(X = cf$cf, MARGIN = 2L, FUN = uw_wrapper))
3: as.data.frame(t(apply(X = cf$cf, MARGIN = 2L, FUN = uw_wrapper)))
2: cbind(as.data.frame(t(apply(X = cf$cf, MARGIN = 2L, FUN = uw_wrapper))), 
       t = (1:ncol(cf$cf)))
1: uwerr.cf(samplecf)
```

PR #227, opened 2020-01-05
Example for uwerr.cf

martin-ueding:
In #226 I raised that the `uwerr.cf` function is broken somehow. This adds an example such that this part of the code is exercised as part of the checks and that further failures will be caught earlier.

### DIFF
--- a/R/cf.R
+++ b/R/cf.R
@@ -563,7 +563,7 @@ jackknife.cf <- function(cf, boot.l = 1) {
 #'         or `cf$icf`, the corresponding row of `uwcf` and/or `uwicf` will contain
 #'         `NA` for all members.
 #' 
-#' @example
+#' @examples
 #' x <- uwerr.cf(samplecf)
 #' summary(x)
 #' plot(x)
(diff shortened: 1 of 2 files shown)


martin-ueding:
I will merge that in *because* it fails such that we get a realistic representation of the state of `master`.

kostrzewa:
it's a typo, also keep in mind that the return value for uwerr.cf has changed, it's now a list (as documented)

martin-ueding:
I just saw that it was this simple, so I have pushed a fix in #228.

martin-ueding:
Wait, so the `plot.uwerr` resolution does not work any more because it is now just a data frame without a class?

kostrzewa:
`plot.uwerr` never worked on the data frame returned by `uwerr.cf`. `plot.uwerr` is for time series of a single value.

kostrzewa:
When I added `uwerr.cf`, I thought about returning a list of `uwerr`, such that `plot.uwerr` can be called on each time slice, but I discarded the idea as I did not fancy looking at `4*time_extent*nr_of_correls` plots :)

kostrzewa:
In retrospect I could have added a summary function which would produce the current output while the default would have been a nested list of `uwerr`...

martin-ueding:
It's fine, I just removed the two lines from the examples. I just extract the data and plot it myself with ggplot anyway. There was `tau_int` instead of `tauint` such that this column was silently dropped. I just fixed that in `master`.

PR #229, opened 2020-01-06
Hankel: resolve edge cases, improve code, add features

urbach:


kostrzewa:
?

urbach:
yes, ?

kostrzewa:
This is suddenly empty...

urbach:
shouldn't be empty...

kostrzewa:
I think the problem was that it was closed accidentally and then showed as empty.

urbach:
shall I simply re-open it...

kostrzewa:
Yes, you have reopened it and now it doesn't show as empty any more. Strange, but probably a performance optimisation on the side of github: if a PR is closed, no need to calculate a diff...

urbach:
cannot install / compile some package...!?

urbach:
@kostrzewa does the code resolve most of the issues you raised?

martin-ueding:
I had to resolve a conflict, I am not sure whether I did so correctly.

urbach:
> I had to resolve a conflict, I am not sure whether I did so correctly.

where was the conflict?

R/hankel.R
```diff
@@ -43,11 +48,17 @@ gevp.hankel.old <- function(cf, t0, deltat = 1, n, N, id=c(1),
 #'    matrix for submatrix.size > 1. \code{element.order=c(1,2,3,4)} leads to a matrix
 #'    \code{matrix(cf[element.order], nrow=2)}.
 #'    Double indexing is allowed.
+#'
+#'
+#' @return
+#' A complex vector of length \code{n + n^2} with firsth the eigenvalues and
```
kostrzewa: would it not make more sense to return a list with named elements?

R/hankel.R
```diff
@@ -118,121 +125,355 @@ gevp.hankel <- function(cf, t0=1, deltat=1, n, N, eps=0.0001, range=c(0,1),
...
 }
 
+#' @title plot_hankel_spectrum
+#'
+#' @description
+#' produces a scatter plot of the complex \eqn{-\log}{-log}
+#' of the eigenvalues produced by the
+#' \link{bootstrap.hankel} method. In addition, produces a
+#' histogramm of all real and positive eigenvalues after computing
+#' \eqn{-\log(ev)/\delta t}{-log(ev)/delta t} in the range
```
kostrzewa: `-log(ev)/\delta t` repeated?

urbach:
> 
> 
> 
> > @@ -43,11 +48,17 @@ gevp.hankel.old <- function(cf, t0, deltat = 1, n, N, id=c(1),
>  #'    matrix for submatrix.size > 1. \code{element.order=c(1,2,3,4)} leads to a matrix
>  #'    \code{matrix(cf[element.order], nrow=2)}.
>  #'    Double indexing is allowed.
> +#'
> +#'
> +#' @return
> +#' A complex vector of length \code{n + n^2} with firsth the eigenvalues and
> 
> would it not make more sense to return a list with named elements?

If you have a neat idea how to save the apply in bootstrap.hankel...

urbach:
> +#' \eqn{-\log(ev)/\delta t}{-log(ev)/delta t} in the range
> 
> `-log(ev)/\delta t` repeated?

One is latex, the other text...

kostrzewa:
> If you have a neat idea how to save the apply in bootstrap.hankel...

Can be implemented as an `lapply` over the rows of the bootstrapped data, thus producing one list element per sample which in turn contains `$evals` and `$evecs` (or simlar). Subsequently, these are turned into the right format in a self-documenting manner:

```r
# simulate application of hankel.gevp to bootstrap samples, extracting three eigenvalues and hence a 3x3 matrix for the eigenvectors
hankel_boot <- lapply(
  X = 1:100,
  FUN = function(r){
    list(evals = sprintf("eval r=%d %s", r, c("a","b","c")),
         evecs = t(outer(X=c("a", "b", "c"), Y=1:3, function(x,y){ sprintf("evec r=%d %s%d", r, x, y) })) 
        )
  }
)

> hankel_boot[[1]]
$evals
[1] "eval r=1 a" "eval r=1 b" "eval r=1 c"

$evecs
     [,1]          [,2]          [,3]         
[1,] "evec r=1 a1" "evec r=1 b1" "evec r=1 c1"
[2,] "evec r=1 a2" "evec r=1 b2" "evec r=1 c2"
[3,] "evec r=1 a3" "evec r=1 b3" "evec r=1 c3"

# extract the bootstrap samples for evals and evecs from the nested list
evals_boot <- t(sapply(hankel_boot, function(x){ x$evals }))
evecs_boot <- t(sapply(hankel_boot, function(x){ x$evecs }))                                                                                      
```

This results in

```
> evals_boot
       [,1]           [,2]           [,3]          
  [1,] "eval r=1 a"   "eval r=1 b"   "eval r=1 c"  
  [2,] "eval r=2 a"   "eval r=2 b"   "eval r=2 c"  
  [3,] "eval r=3 a"   "eval r=3 b"   "eval r=3 c"  
  [4,] "eval r=4 a"   "eval r=4 b"   "eval r=4 c"  
  [5,] "eval r=5 a"   "eval r=5 b"   "eval r=5 c"  
```

and 

```
> evecs_boot
       [,1]            [,2]            [,3]            [,4]           
  [1,] "evec r=1 a1"   "evec r=1 a2"   "evec r=1 a3"   "evec r=1 b1"  
  [2,] "evec r=2 a1"   "evec r=2 a2"   "evec r=2 a3"   "evec r=2 b1"  
  [3,] "evec r=3 a1"   "evec r=3 a2"   "evec r=3 a3"   "evec r=3 b1"  
  [4,] "evec r=4 a1"   "evec r=4 a2"   "evec r=4 a3"   "evec r=4 b1"  
  [5,] "evec r=5 a1"   "evec r=5 a2"   "evec r=5 a3"   "evec r=5 b1"  
  [6,] "evec r=6 a1"   "evec r=6 a2"   "evec r=6 a3"   "evec r=6 b1"  
  [7,] "evec r=7 a1"   "evec r=7 a2"   "evec r=7 a3"   "evec r=7 b1"  
  [8,] "evec r=8 a1"   "evec r=8 a2"   "evec r=8 a3"   "evec r=8 b1"  
  [9,] "evec r=9 a1"   "evec r=9 a2"   "evec r=9 a3"   "evec r=9 b1"  
```

which should correspond to the desired dimensioning, right?

I don't care too much since `hankel.gevp` is not user-facing, but I always find it somewhat unintuitive to have a return value which mixes two different things in a simple vector.

martin-ueding:
The merge conflict was in the function signature of `bootstrap.hankel`. In the `master` branch it has this form:

```r
bootstrap.hankel <- function(cf, t0, n=2, N = cf$Time/2+1, id=c(1), range=c(0,1), eps=0.001) {
```

In this pull request it has

```r
bootstrap.hankel <- function(cf, t0, n=2, N = cf$Time/2+1, id=c(1)) {
```

Often parameters are added, but it seems that here these parameters have been removed. So I resolved the merge conflict such that the latter form is now used.

urbach:
> The merge conflict was in the function signature of `bootstrap.hankel`. In the `master` branch it has this form:
> 
> ```r
> bootstrap.hankel <- function(cf, t0, n=2, N = cf$Time/2+1, id=c(1), range=c(0,1), eps=0.001) {
> ```
> 
> In this pull request it has
> 
> ```r
> bootstrap.hankel <- function(cf, t0, n=2, N = cf$Time/2+1, id=c(1)) {
> ```
> 
> Often parameters are added, but it seems that here these parameters have been removed. So I resolved the merge conflict such that the latter form is now used.

that's correct!

urbach:
@kostrzewa so, is it worth re-writing the interface? And what happens if you `X` is not a vector, but an array like in the `bootstrap.hankel` method?

kostrzewa:
> And what happens if you `X` is not a vector, but an array like in the `bootstrap.hankel` method?

I don't quite understand the question as we're talking about `hankel.gevp`, which takes a vector as input and currently produces a vector as output which actually contains eigenvalues and eigenvectors linearised into a single vector, without a definition of the ordering of the eigenvector components (the ordering is obvious in the end, but it's still undefined as far as the interface is concerned).

The sample code that I gave simulates what happens in `bootstrap.hankel` when `hankel.gevp` is called on the bootstrap samples and how one can replace `apply` with `lapply`. `X` (i.e. `1:100` in the example) are the row indices of `cf.tsboot`:

```r
hankel_boot <- lapply(
  X=1:nrow(cf.tsboot),
  FUN=function(r){ 
     ... # do stuff for bootstrap sample `r`
     return( list(evals = ..., evecs = ...) )
      }
)
```

This would produce one list element per bootstrap sample, each of which contains `evals` and `evecs` for that bootstrap sample (self-documenting return value). The two `sapply` calls then bring the result into array form. For the eigenvectors this is not trivial, of course, as they should eventually become objects with three, rather than two, indices.
 
> @kostrzewa so, is it worth re-writing the interface? 

No, in this case, as `hankel.gevp` is not user-facing, I don't think it's worth the effort.

urbach:
but then this is more or less ready from my point of view.

PR #230, opened 2020-01-06
Add prior docu within bootstrap.nlsfit.R

NikSchlage:
Add prior documentatoin within bootstrap.nlsfit.R


urbach:
> martin-ueding requested changes on this pull request.
> 
> The documentation itself is very nice and precise, thank you very much!
> 
> The file `man/bootstrap.nlsfit.Rd` and `man/simple.nlsfit.Rd` are
> not updated in thie pull request, though. You need to run the commit
> hook or `devtools::document()` and also add these changes to the
> pull request.

But only the two files... ;)

kostrzewa:
Ah, the fun of different roxygen versions! Maybe it's time to really throw out all the generated `.Rd` files and require users who fetch from the repo to install roxygen and devtools... ?

NikSchlage:
Thank you for your feedback. I think that I just updated and pushed all man files... Is there something wrong with that?

martin-ueding:
> Maybe it's time to really throw out all the generated .Rd files and require users who fetch from the repo to install roxygen and devtools... ?

Please!

kostrzewa:
> > Maybe it's time to really throw out all the generated .Rd files and require users who fetch from the repo to install roxygen and devtools... ?
> 
> Please!

@urbach, it's your call

urbach:
> > > Maybe it's time to really throw out all the generated .Rd files and require users who fetch from the repo to install roxygen and devtools... ?
> > 
> > Please!
> 
> @urbach, it's your call

okay, if you want go ahead. I just don't have time to do this... And
who has?

kostrzewa:
> Thank you for your feedback. I think that I just updated and pushed all man files... Is there something wrong with that?

Well, you only modified two of them, the rest were modified as a side-effect because you are likely using a different roxygen version than the one that was used to generate them previously. This means that

1) your commit is mislabeled: you've updated all `.Rd` files, not just two of them
2) your commit touches paths that it should't touch (not your fault though, just a side-effect of having generated files in the repo)

As this happens quite frequently, we have quite a few commits in the repo which make substantial (but irrelevant) changes to lots of files. When trying to understand commit history and/or trying to track how a bug originated, all this noise really gets in the way...

martin-ueding:
> And who has?

I'm more than happy to nuke the generated files 💣

kostrzewa:
Please go ahead

NikSchlage:
> > Thank you for your feedback. I think that I just updated and pushed all man files... Is there something wrong with that?
> 
> Well, you only modified two of them, the rest were modified as a side-effect because you are likely using a different roxygen version than the one that was used to generate them previously. This means that
> 
>     1. your commit is mislabeled: you've updated all `.Rd` files, not just two of them
> 
>     2. your commit touches paths that it should't touch (not your fault though, just a side-effect of having generated files in the repo)
> 
> 
> As this happens quite frequently, we have quite a few commits in the repo which make substantial (but irrelevant) changes to lots of files. When trying to understand commit history and/or trying to track how a bug originated, all this noise really gets in the way...

I'll update roxygen2 and fix that.

kostrzewa:
> I'll update roxygen2 and fix that.

No need, @martin-ueding is removing all generated files such that this kind of thing doesn't happen in the future, see #232

NikSchlage:
Ok, that's nice.

PR #231, opened 2020-01-07
Add a single-constant model

martin-ueding:
Just a model which is the usual “single” model with a constant offset.


martin-ueding:
Here is a demonstration: [single_constant_model.pdf](https://github.com/HISKP-LQCD/hadron/files/4030423/single_constant_model.pdf)

kostrzewa:
I'll hold off on the merge the generated files are removed.

PR #235, opened 2020-01-07
Remove rd files

urbach:
steps towards removing all Rd files


kostrzewa:
This reformats quite a bit of existing roxygen docs, why?

urbach:
> This reformats quite a bit of existing roxygen docs, why?

uh, okay, that was not intended. So the mapping is not
bijective... Then I'll need to do it again.

urbach:
see #236

PR #237, opened 2020-01-12
GEVP vignette

urbach:
A vignette to explain how to use the GEVP.


martin-ueding:
It seems that this suffers from the biber issue as well as the Hankel vignette.

urbach:
> It seems that this suffers from the biber issue as well as the Hankel vignette.

Why did that stop working?

kostrzewa:
> Why did that stop working?

One of the annoyances of using a third-party CI service instead of having one's own infrastructure...  As I said elsewhere, there are incompatibilities between the texlive installed on the travis instance and those added in the course of setting up our package.

```
ERROR - Error: Found biblatex control file version 3.3, expected version 3.7.
This means that your biber (2.14) and biblatex (3.7) versions are incompatible.
```

Some people seem to resolve this kind of annoyance by setting up their own docker image which contains all dependencies and then run their CI in that docker image...

kostrzewa:
It might also be possible to simply install all of texlive as `apt_packages`, not sure if that'll help though.

kostrzewa:
https://ljvmiranda921.github.io/notebook/2018/04/23/postmortem-shift-to-docker/

urbach:
hmpf, not sure whether I want to go through this. Should I just remove
the references for the time being?

kostrzewa:
See #238 for a possible solution... not sure that I like it though!

vignettes/gevp.Rmd
```diff
@@ -19,13 +19,13 @@ library(knitr)
 library(hadron)
 ```
 
-# Data Type \texttt{cf} for Correlation Functions
+# Data Type `cf` for Correlation Functions
 
-\texttt{hadron} provides a data type or better class for correlation
-functions and correlation matrices, which is called \texttt{cf}. There
+`hadron` provides a data type or better class for correlation
```
kostrzewa: does this render properly in the end?

kostrzewa: Sorry, not here, but when used within roxygen-generated docs?

martin-ueding: It is a markdown document, so using backticks is more proper than using LaTeX code. So for instance this won't work when it is rendered as HTML.

martin-ueding:
I cannot find the `gevp.bib` file that is supposed to be present here. So I cannot build the vignette locally. Where is it? And why does it work on Travis CI now?

vignettes/gevp.Rmd
martin-ueding: So far I have the impression that Roxygen properly does Markdown.

kostrzewa:
>  And why does it work on Travis CI now?

Yeah, that's what I was wondering about too. I scanned the travis logs and could not find any mention of the vignettes even being built...

martin-ueding:
Somewhere in the latest log it says this:

    $ export R_CHECK_ARGS="--no-build-vignettes --no-manual --timings"

I am not sure why this option is there, we had failing vignettes all the time?

martin-ueding:
I'd say we merge this and then deal with all the vignette-on-Travis-CI issues in a separate PR?

urbach:
> I cannot find the `gevp.bib` file that is supposed to be present here. So I cannot build the vignette locally. Where is it? And why does it work on Travis CI now?

no references yet, so no bib file...

urbach:
> I'd say we merge this and then deal with all the vignette-on-Travis-CI issues in a separate PR?

yes...

PR #238, opened 2020-01-15
Gevp vignette texlive

kostrzewa:
This seems to fix the texlive issue, although I'm not 100% sure it's a good solution...


martin-ueding:
It is just on Travis CI, so as long as it works, it is okay for now.

kostrzewa:
We just need to keep it in mind as it installs R packages independently of the rest of `.travis.yml` and I think it might interfere in the future.

PR #241, opened 2020-01-20
Treat warnings as errors on Travis CI

martin-ueding:
Now we will get serious about a stable CRAN submission and enforce that
no new warnings will get in. We likely need to stabilize this branch
first, though.


martin-ueding:
Okay, so this has passed, and `master` will pass too therefore. So now we locked in a warning free code (hopefully). This is going to be fun! :smiling_imp:

PR #245, opened 2020-01-28
Further exports

kostrzewa:
some more exports which are required in client code

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: hadron
-Version: 3.0.0
+Version: 3.0.1
 Date: 2020-01-16
 Title: package to extract hadronic quantities
 Authors@R: c(person("Carsten", "Urbach", email = "urbach@hiskp.uni-bonn.de",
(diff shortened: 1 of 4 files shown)


kostrzewa:
I bump the version to make the client code aware.

martin-ueding:
I don't get why the unit tests fail for the PR run from these changes. Guess the tests are super flaky because there aren't actually two states of data in there.

kostrzewa:
> I don't get why the unit tests fail for the PR run from these changes. Guess the tests are super flaky because there aren't actually two states of data in there.

Yeah, I have no idea. At first push, it was also only the 'PR' travis run

https://travis-ci.org/HISKP-LQCD/hadron/builds/642865972?utm_source=github_status&utm_medium=notification

which failed while the 'branch' travis run passed just fine...

https://travis-ci.org/HISKP-LQCD/hadron/builds/642865895?utm_source=github_status&utm_medium=notification

I guess the latter does not run all tests though.

kostrzewa:
It must be some problem on Travis as now both test runs have passed (and I basically didn't change anything other than merging #244 via `master`)

kostrzewa:
Travis ...

kostrzewa:
ping

PR #247, opened 2020-01-29
Repair tex-utils.R

NikSchlage:
If fit parameters and corresponding errors get really small, problems arise within `tex-utils.R` in
`while(round(10^N*abs(x)) < threshold)`


j-ostmeyer:
The last change mainly affected lines @kostrzewa modified last, where NAs were set to zero. This is plainly wrong, so we decided to print NAs as they are.

PR #249, opened 2020-02-11
improve documentation for analysis_gradient_flow and add flag to cont…

kostrzewa:
…rol the plot size


martin-ueding:
I have deleted the branch here on GitHub, but you can just push again if you were not finished with this.

PR #251, opened 2020-02-13
add test for super tricky tex.catwitherror formatting issue

kostrzewa:
This is a really annoying formatting issue that I've just encountered in production. The fix would be to first round the error to the next highest digit and then to base the rounding of the value on that...


kostrzewa:
I'm not sure if we want to invest the effort to fix this... in the use-case in question, it would be more appropriate to use `digits=2`, but the plots that I use this in are quite busy, so `digits=1` was used to save space.

j-ostmeyer:
I tried out the fix you suggested. It is not complicated and works in principle. Yet, it crashes another test and this is why:
```
> round(12.345, digits=2)
[1] 12.35
```
BUT
```
> round(1234.5)
[1] 1234
```

This leads to
```
test_tex_catwitherror.R:23: failure: same_error
tex.catwitherror(12.345, 12.345, digits = 4, with.dollar = FALSE) not equal to "12.35(12.35)".
1/1 mismatches
x[1]: "12.35(12.34)"
y[1]: "12.35(12.35)"
```

I just found that funny. We can use the fix or we can let everything as it is. I don't mind.

tests/testthat/test_tex_catwitherror.R
```diff
@@ -20,7 +20,7 @@ test_that('very_small_number', {
 
 test_that('same_error', {
     expect_equal(tex.catwitherror(0.00123, 0.00123, digits = 4, with.dollar = FALSE), '0.001230(1230)')
-    expect_equal(tex.catwitherror(12.345, 12.345, digits = 4, with.dollar = FALSE), '12.35(12.35)')
```
kostrzewa, line 19: why this change?

j-ostmeyer: Because the test fails otherwise. See above.

kostrzewa: Sorry, yes I missed that somehow (I was only looking at the changes). I think it's okay to round down .5, as .500000001 is rounded up.

R/tex-utils.R
```diff
@@ -107,6 +107,8 @@ tex.catwitherror <- function(x, dx, digits=1, with.dollar=TRUE, human.readable=T
     }
 
     if(!is.na(err) && err > 0) {
+      N <- absolute.number.digits(err, digits)
```
kostrzewa, line 110: could we add a comment here that also reference the `borderline_error` test? Otherwise the logic seems a little ad hoc and smells like an accidental drop of precision.

j-ostmeyer: Sure.

PR #254, opened 2020-02-20
add function for computing GL FSE correction factors

kostrzewa:


kostrzewa:
I guess the documentation needs some formatting love...

kostrzewa:
Since I went with just returning the correction factors, `K_fps` and `K_Mps` in a data frame, rather than the infinite-volume values of `M_ps` and `f_ps`, I've adjusted the documentation and removed the now unnecessary parameters for the mass and decay constant.

kostrzewa:
> Hi Bartek,
> Am I right that a factor of 2 is missing from the correction to M_ps ?
> K_Mps = (1+ 1/(2_N_f)_xi_g1_tilde), where Nf=2 so K_Mps = (1+ 0.25_xi*g1_tilde),

We work in the normalisation where `f_pi ~ 130.4 MeV` and `\xi = (M_pi / 4 \pi f_pi)^2`, so the formulas should be correct as implemented for this convention. When `F_pi = f_pi/sqrt(2)` is used instead, the factors change of course.

pittlerf:
> > Hi Bartek,
> > Am I right that a factor of 2 is missing from the correction to M_ps ?
> > K_Mps = (1+ 1/(2_N_f)_xi_g1_tilde), where Nf=2 so K_Mps = (1+ 0.25_xi*g1_tilde),
> 
> We work in the normalisation where `f_pi ~ 130.4 MeV` and `\xi = (M_pi / 4 \pi f_pi)^2`, so the formulas should be correct as implemented for this convention. When `F_pi = f_pi/sqrt(2)` is used instead, the factors change of course.

Thanks, I have just realized also that.

kostrzewa:
I still need to fix the issues with the documentation, so don't merge this yet please...

PR #255, opened 2020-02-20
another borderline error which results is unexpected behaviour, this …

kostrzewa:
…time it produces an exponential


kostrzewa:
```
     expect_equal(tex.catwitherror(0.031921636680, 0.000098, digits = 1, with.dollar=FALSE), '0.0319(1)') 
```

we get: `0.03192(1e-04)` which, while correct, is terrible.

martin-ueding:
I don't think that `0.03192(1e-04)` is correct. We rather want `0.03192(1)`.

kostrzewa:
> I don't think that 0.03192(1e-04) is correct. We rather want 0.03192(1).

I think we rather want `0.0319(1)`, no? We have:

```
val = 0.031922
err = 0.000098
```

which I would expect to be rounded to 

```
val = 0.0319
err = 0.0001
```

to obtain `0.0319(1)`. In that sense I think the `err = 1e-4` that comes out with the present logic is correct, but the scientific notation is clunky and the value should be rounded to the same level as the error was. I don't know how easy it will be to make this happen without destroying wanted behaviour in other edge cases...

martin-ueding:
I messed up the digits. All I wanted to say is that having a scientific exponential in the parentheses makes it incorrect because the overlapping rule of the digits would then lead to an error four orders of magnitude smaller if interpreted strictly.

j-ostmeyer:
Can be fixed by the same change as in #251 .

PR #265, opened 2020-04-11
Convert logical masks to integer masks

martin-ueding:
The documentation promises that logical masks are fine, but they are not when one uses xy-errors. This should fix it.


kostrzewa:
lgtm

kostrzewa:
just the two locations though?

martin-ueding:
I believe only these two locations are needed as all the parametric entry points just pass `...` to the `bootstrap.nlsfit` function and therefore we do not need do more.

PR #268, opened 2020-04-22
Hankel

urbach:
this should be hopefully the final part of the hankel method:

- additional parameter Delta
- a bug in the physical time versus R index was fixed **which will influence fit ranges** in already existing project bei a shift by 1
- `/.cf` has been modified to also work for principal correlators (without original data)


martin-ueding:
I've just tried this out in my analysis and it breaks with this error:

```
Error in 1:(N - 1 - reftime - 2 * (n/submatrix.size - 1) * Delta) : 
  argument of length 0
Called from: hankel2cf(hankel = hankel, id = id, range = range, eps = eps, 
    sort.type = sort.type, sort.t0 = sort.t0)
```

My calling code is this here:

```r
range <- -c(mass_val + 1000 * mass_err, mass_val - 1000 * mass_err)

pc_hankel_running_effmass <- hankel2effectivemass(
        value$pc_hankel_running,
        id = 1,
        range = exp(range)
        )
```

And that Hankel object comes from here:

```r
pc_hankel_running <- bootstrap.hankel(cf = value$pc_ratio, t0 = param$hankel_t0, n = 2, t0fixed = FALSE)
```

I am not sure what to make of this error, but it breaks my analysis.

urbach:
> I've just tried this out in my analysis and it breaks with this error:
> 
> ```
> Error in 1:(N - 1 - reftime - 2 * (n/submatrix.size - 1) * Delta) : 
>   argument of length 0
> Called from: hankel2cf(hankel = hankel, id = id, range = range, eps = eps, 
>     sort.type = sort.type, sort.t0 = sort.t0)
> ```
> 
> My calling code is this here:
> 
> ```r
> range <- -c(mass_val + 1000 * mass_err, mass_val - 1000 * mass_err)
> 
> pc_hankel_running_effmass <- hankel2effectivemass(
>         value$pc_hankel_running,
>         id = 1,
>         range = exp(range)
>         )
> ```
> 
> And that Hankel object comes from here:
> 
> ```r
> pc_hankel_running <- bootstrap.hankel(cf = value$pc_ratio, t0 = param$hankel_t0, n = 2, t0fixed = FALSE)
> ```
> 
> I am not sure what to make of this error, but it breaks my analysis.

that's a bit hard to debug for me in this way. Do you have an object on which I can try myself?

martin-ueding:
This is the `value` object that fails. It contains the Hankel object that I try to put into the `hankel2effectivemass` function.
[value.rds.gz](https://github.com/HISKP-LQCD/hadron/files/4574302/value.rds.gz)

urbach:
hmm, I tried `load("value.rds")` and got this:

    Error in load("value.rds") : 
      bad restore file magic number (file may be corrupted) -- no data loaded
    In addition: Warning message:
    file ‘value.rds’ has magic number 'X'
      Use of save versions prior to 2 is deprecated

martin-ueding:
I stored that with writeRDS, you have to do `value <- readRDS('value.rds')` to load it.

urbach:
`Delta` and `submatrix.size` are somehow not part of `pc_hankel_running`. I'll investigate...

urbach:
did you use the most up-to-date `bootstrap.hankel` ? Because `Delta` and `submatrix.size` are explicitly set for the returned list.

martin-ueding:
I have removed all the intermediate results that were generated with the Hankel code and re-run the analysis. It seems that all the fits run through now.

PR #271, opened 2020-04-30
ignore some more files in R CMD build

urbach:
needed to remove some more NOTES and WARNINGS for CRAN submission

and reduce the run-time of examples significantly


kostrzewa:
double entry of `travis...yml` in `.Rbuildignore`. Also, should that include `travis_install_texlive.R` (it could of course also be removed now as it does not seem to be required any more)

Issue #256, opened 2020-02-23
Correlation functions from extract.obs cannot be bootstrapped


pittlerf:
I try to bootstrap correlation functions extracted from online measurements data using extract.obs,
and I got the following error:
```
bootstrap.cf(ppcorrelator, boot.R=1500, boot.l=1)
Error in apply(x, MARGIN = 2L, FUN = mean) : 
  dim(X) must have a positive length
```
I think the problem is that bootstrap.cf still think that there is an icf field in the correlation function, because it was instantiated with 
```
ret <- cf_orig(ret, cf = cf, icf = NA)
```
and 
```
> has_icf(ppcorrelator)
[1] TRUE
``` 
The solution could be 
```
ret <- cf_orig(ret,cf=cf,icf=NULL)
```

urbach:
well, `extract.obs` is missing the corresponding entry

PR #274, opened 2020-04-30
fix for #256

urbach:


urbach:
hmm, problem is package `xfun`

Error: package ‘xfun’ was installed before R 4.0.0: please re-install it

Execution halted

ERROR: lazy loading failed for package ‘tinytex’

* removing ‘/home/travis/R/Library/tinytex’

PR #275, opened 2020-04-30
fix travis

urbach:


urbach:
hmm, why did it work now? I'm lost with travis...

urbach:
this was now with R 3.6.3, so maybe that helped?

urbach:
it seems to be a problem with the new R version (4.0.X). Maybe its a good idea to stick to 3.6.X for the time being.

PR #277, opened 2020-05-04
use skip_on_os to disable the 5 failing test for windoof

urbach:
this is an alternative to #276, which way to we want to take?


martin-ueding:
I like this approach much better as it allows us to still have some tests running there.

PR #278, opened 2020-05-05
fix first complains during CRAN submission

urbach:
in the process of the CRAN submission there are popping up points and changes I'll collect here


urbach:
I have basically everything implemented now, apart from the author ship thing...

kostrzewa:
Would be happy to merge this, but it will conflict with the multiplication function from #280 It still has a "do not merge" tag.

kostrzewa:
- [ ] Thanks, please use only undirected quotations marks in your Description text ('boot').
- [ ] Please add '()' behind all function names (e.g. bootstrap.nlsfit()  without quotation marks) in your Description text.
- [ ] Please replace cat() by message() or warning() in your functions (except for print() and summary() functions). Messages and warnings can be suppressed if needed.
- [ ] Please do not change the .Random.seed or the global environment in your functions.
- [ ] You are changing the user's par() settings in your functions. Please ensure with an immediate call of on.exit() that the settings are reset. E.g.
- [ ] Please do not change the user's warning options in your functions. If you really have to suppress warnings, please use suppressWarnings().

kostrzewa:
The change of `par()` goes on me, at least in the exported functions from `analysisonline.R` and `timeseries.R`. I'll take care of it.

kostrzewa:
Some of these remaining issues are a lot of very boring work...

urbach:
Yes, I know...
-- 
CU Mobil, www.carsten-urbach.de

urbach:
why do we actually need `restore_seed`?

urbach:
maybe we should use the `setRNG` package?

kostrzewa:
> why do we actually need `restore_seed`?

One could of course argue that using a function which performs bootstrapping may legitimately have the side-effect of modifying the global RNG state (just like `sample` and other functions do). On the other hand, having a function without global side effects is probably a nice thing in general. Not sure what's better... If we remove `restore_seed`, then I would argue for removing the default value for the seed in all bootstrapping functions which take one in order to make it apparent that the function has global side effects.

urbach:
> > why do we actually need `restore_seed`?
> 
> One could of course argue that using a function which performs bootstrapping may legitimately have the side-effect of modifying the global RNG state (just like `sample` and other functions do). On the other hand, having a function without global side effects is probably a nice thing in general. Not sure what's better... If we remove `restore_seed`, then I would argue for removing the default value for the seed in all bootstrapping functions which take one in order to make it apparent that the function has global side effects.

no, I think we need to be able to reproduce bootstrap samples from a
saved seed. And for this we need `restore_seed`, as far as I can see.

I'd like to ask, but will we get any answer?

kostrzewa:
I don't think that we need `restore_seed` to have reproducibility we don't use the RNG state for this, only the integer seed. As far as I can tell, `restore_seed` is used only to avoid global side-effects.

urbach:
> I don't think that we need `restore_seed` to have reproducibility we don't use the RNG state for this, only the integer seed. As far as I can tell, `restore_seed` is used only to avoid global side-effects.

I'm not sure I fully agree. I would think we need to be able to use
e.g. `bootstrap.cf` on different correlation functions with the same
seed to preserve correlations. But, in my opinion, it would be very
bad if the random number generator is in the same state after every
call to `bootstrap.cf`. And I want to set the seed in `bootstrap.cf`,
because I want to store the seed in the returned object such that we
can reproduce this later on. If we are not supposed to touch
`.Random.seed`, we could also do the following:

    nseed <- sample.int(1, 138998)
    set.seed(seed)
    ## bootstrap here
    ##
    set.seed(nseed)

Couldn't we?

This is still reproducable and avoids the problem I described
above. Many functions in R have the global side effect to change the
random number state. This alone is not a problem for me.

kostrzewa:
> I'm not sure I fully agree. I would think we need to be able to use e.g. `bootstrap.cf` on different correlation functions with the same seed to preserve correlations.

Which is exactly what we do.

> But, in my opinion, it would be very bad if the random number generator is in the same state after every call to `bootstrap.cf`.

But this is **exactly** the current situation now (with the subtlety that the RNG state is only reset if a state already existed). The seed is set, the RNG runs (via bootstrap.cf, for example) and the state of the RNG is reset afterwards. 

> And I want to set the seed in `bootstrap.cf`, because I want to store the seed in the returned object such that we can reproduce this later on.

Of course, that's exactly what we do, but for this we don't need to touch `.Random.seed`, it's sufficient to call `set.seed(seed)` and store `stored_seed <- seed`. When we need to use the same random numbers again, we call `set.seed(stored_seed)`, which is the first part of what we do now.

**However**, in addition to setting the seed, we also fully reset the state of the RNG, such that functions like `bootstrap.cf` actually don't have global side effects (unless the state was not previously set). One can argue for or against this. On the one hand, global side effects are generally bad, on the other hand, as you say, having the RNG in the initial state after a `bootstrap.cf` call is also not ideal.

> Many functions in R have the global side effect to change the random number state. This alone is not a problem for me.

Sure, as I said, we can decide to *stop* resetting the RNG state and then we don't need to touch `.Random.seed` at all. This brings with it the complication that one has to be aware of the fact that `bootstrap.cf` sets the seed, generates RNs and then leaves the RNG in a given state, which will always be the same, as you say. However, that's not very different from the current situation (up to the subtlety mentioned above).
     
> ```
>nseed <- sample.int(1, 138998)
>set.seed(seed)
>## bootstrap here
>##
>set.seed(nseed)
>```

Yeah, if we accept global side effects, this is a solution which will work. There is only one edge-case that I can think of which involves the seed being externally set before this call, which would always produce the same `nseed`, such that the RNG would always end up in the same state. However, that's an external problem.

urbach:
> > I'm not sure I fully agree. I would think we need to be able to use e.g. `bootstrap.cf` on different correlation functions with the same seed to preserve correlations.
> 
> Which is exactly what we do.

yes, I know.

> > But, in my opinion, it would be very bad if the random number generator is in the same state after every call to `bootstrap.cf`.
> 
> But this is **exactly** the current situation now (with the subtlety
> that the RNG state is only reset if a state already existed). The
> seed is set, the RNG runs (via bootstrap.cf, for example) and the
> state of the RNG is reset afterwards.

No, maybe I don't understand, but currently we can have

    bootstrap.cf(data, seed=1)
    runif(1)
    bootstrap.cf(other data, seed=1)
    runif(1)

and the two random numbers will be different. If we don't restore the
state, the random numbers will be identical, will they not?

> > And I want to set the seed in `bootstrap.cf`, because I want to store the seed in the returned object such that we can reproduce this later on.
> 
> Of course, that's exactly what we do, but for this we don't need to
> touch `.Random.seed`, it's sufficient to call `set.seed(seed)` and
> store `stored_seed <- seed`. When we need to use the same random
> numbers again, we call `set.seed(stored_seed)`, which is the first
> part of what we do now.

yes, I know. But it speaks against not having seed as a parameter, as
you proposed.

> Sure, as I said, we can decide to *stop* resetting the RNG state and
>      then we don't need to touch `.Random.seed` at all. This brings
>      with it the complication that one has to be aware of the fact
>      that `bootstrap.cf` sets the seed, generates RNs and then
>      leaves the RNG in a given state, which will always be the same,
>      as you say. However, that's not very different from the current
>      situation (up to the subtlety mentioned above).

This I would not like, if I understand correctly, see above.


> > ```
> >nseed <- sample.int(1, 138998)
> >set.seed(seed)
> >## bootstrap here
> >##
> >set.seed(nseed)
> >```
> 
> Yeah, if we accept global side effects, this is a solution which
> >will work. There is only one edge-case that I can think of which
> >involves the seed being externally set before this call, which
> >would always produce the same `nseed`, such that the RNG would
> >always end up in the same state. However, that's an external
> >problem.

Is the global side effect problem anywhere? And the edge-case is for
me expected behaviour, isn't it?

kostrzewa:
> ```
> bootstrap.cf(data, seed=1)
> runif(1)
> bootstrap.cf(other data, seed=1)
> runif(1)
> ```

Yes, in this sense your are absolutely right. It's as if `bootstrap.cf` had never run in between and from my point of view is the perfect situation and it is rather irksome that the CRAN-people have a problem with it...

> yes, I know. But it speaks against not having seed as a parameter, as you proposed.

Sorry, I must have been unclear: I wanted to remove the **default value only**, of course our bootstrapping functions have to have a seed parameter.

I think in order to appease the CRAN-gods we probably have to implement your proposal. Maybe I misunderstand what the `setRNG` package does, but it seems to me that using it would require significant modifications to most of our functions, wrapping stuff in the `simulation` objects provided by the package.

kostrzewa:
> Is the global side effect problem anywhere?

Hard to say. If the RNG state is modified by a call which doesn't make it perfectly obvious that it does so, I would say that might cause issues down the line. For us here it definitely won't cause any problems.

> And the edge-case is for me expected behaviour, isn't it?

Agreed.

kostrzewa:
> This I would not like, if I understand correctly, see above.

Yes, I agree.

urbach:
when is a good time to merge this?

urbach:
okay, here the status, any contribution/comment welcome

- [ ] Please write TRUE and FALSE instead of T and F.
- [ ] You have examples for unexported functions which cannot run in this way. (I still don't know what she means here)
- [ ] Please add small files needed for the examples or vignette in the inst/extdata ...
- [ ] \dontrun{} should only be used if the example really cannot be executed

Anyone concerning the second. 
For the third I added some things, but how much more are we expected to add?
`\dontrun` is now only in IO routines and `tikz.init`. I don't see how to get around those!?

kostrzewa:
The `dontrun` discussion is idiotic to be honest. The CRAN documentation does not list any limitations on what can or cannot be executable examples, as long as those that are executable actually are:

> Examples are not only useful for documentation purposes, but also provide test code used for diagnostic checking of R code. By default, text inside \examples{} will be displayed in the output of the help page and run by example() and by R CMD check. ***You can use \dontrun{} for text that should only be shown***, but not run, and \dontshow{} for extra commands for testing that should not be shown to users, but will be run by example(). (Previously this was called \testonly, and that is still accepted.)

urbach:
> The `dontrun` discussion is idiotic to be honest. The CRAN documentation does not list any limitations on what can or cannot be executable examples, as long as those that are executable actually are:
> 
> > Examples are not only useful for documentation purposes, but also provide test code used for diagnostic checking of R code. By default, text inside \examples{} will be displayed in the output of the help page and run by example() and by R CMD check. ***You can use \dontrun{} for text that should only be shown***, but not run, and \dontshow{} for extra commands for testing that should not be shown to users, but will be run by example(). (Previously this was called \testonly, and that is still accepted.) 

well, I think we should try more or less in the current version.

Does someone see non-exported functions with examples?

urbach:
so, I think I'd try again...!?

kostrzewa:
There are `@examples` without `@export` in `R/hadron-package.R`. Is that supposed to be like that?

kostrzewa:
Other than that I was not able to find anything. Thanks a lot for all the work that you've invested in getting this CRAN-ready.

urbach:
> There are `@examples` without `@export` in `R/hadron-package.R`. Is that supposed to be like that?

yes, I know about those. But this is for data, does it need exporting?
The examples are, of course, also rather pointless, we can just remove
them...

urbach:
> Other than that I was not able to find anything. Thanks a lot for all the work that you've invested in getting this CRAN-ready.

thanks!

urbach:
> > There are `@examples` without `@export` in `R/hadron-package.R`. Is that supposed to be like that?
> 
> yes, I know about those. But this is for data, does it need exporting?
> The examples are, of course, also rather pointless, we can just remove
> them...
> 

"Datasets: all datasets are publicly available. They exist outside of
the package namespace and should not be exported."


The `cars` dataset has an example, too, even if a bit more helpful one.

urbach:
now that `hadron` is soon on CRAN, shall we merge this in?

kostrzewa:
The checks are not passing or don't run.

kostrzewa:
Beyond that, I think it should be merged. If any analyses break as a result, those just have to be fixed...

urbach:
I'm quite sure that it passes all checks, I have run those so many times... I don't know why they don't run here, how can I trigger them to run again?

urbach:
I don't understand, why the checks don't run here. I'd merge this anyhow if there is no protest...

kostrzewa:
Yes, please go ahead. We'll see if the checks run on master...

PR #281, opened 2020-05-20
bootstrap.effectivemass: Do not rely on original data

martin-ueding:
The `bootstrap.effectivemass` function works on a bootstrapped correlator. The data member `cf` belongs to the `cf_orig` mixin and is not present in general cases. Therefore this will fail. But we can just use `cf0` which has to be present.


R/effectivemass.R
```diff
@@ -217,10 +217,7 @@ bootstrap.effectivemass <- function(cf, type="solve") {
   stopifnot(inherits(cf, 'cf_boot'))
 
   ## number of measurements
-  N <- length(cf$cf[,1])
-  if(is.null(cf$cf)) {
-    N <- cf$N
-  }
+  N <- length(cf$cf0)
```
kostrzewa, line 220: this cannot be correct, the number of measurements is not the length of `cf0`

kostrzewa: The order of operations on the LHS is simply wrong and probably the right thing to do would be to inherit `cf$N` to begin with.

martin-ueding: Oh, you are right. It is the number of measurements and not the number of time slices.

But why does the effective mass need to know the number of measurements anyway?

martin-ueding: Oh, it just preserves it. I think it just does not belong in there, just as the original correlator should not be in it. At least we have “effectivemass *has* cf” here, unlike “matrixfit *is* cf”.

urbach: but now `N` is not defined anymore? `N` is still part of the returned `list`, isn't it?

martin-ueding:
I see that we have two options: Remove `N` completely from the effective mass, or make `N` a requirement of the `cf_meta` mixin and make sure that it is present *everywhere*.

martin-ueding:
I have removed this now completely as the `cf_meta` does not have this as a specification.

PR #296, opened 2020-08-03
Create the configure during build time

### LINKED ISSUE #286: configure is a generated file and should not be checked in
I just ran `autoreconf` and got changes in git. I think that the `configure` should not be tracked in git.

martin-ueding:
Closes #286.

I propose the needed changes such that `configure` is created automatically.


martin-ueding:
I think since we have the package on CRAN, we have the separate releases like the GNU Autotools model uses. And then in the development process we can assume that developers have all the tools needed.

The edge case would be an end-user wanting to install a development version on a machine without GNU Autotools installed. Does that really happen?

PR #301, opened 2020-08-04
Let Travis CI fail on warnings

martin-ueding:
Apparently there is no command line flag, so we need to parse the output for `WARNING` and then fail.


martin-ueding:
The PR check is fine, but the push check has failed due to a timeout in the GPG key retrieval. I would say that the checks are fine and that we don't have any warnings in the build. There are only two `NOTE`, but they are just that the size of our package is quite large. The second is that if we would install ghostscript, it would compress the PDFs a bit. But there is no need for that during testing.

So I think that this can be merged.

PR #309, opened 2022-06-21
Cf adds

NikSchlage:
Add functions to

- exponentiate cf objects
- assign numeric values to entries of cf objects
- compute the FHT ratio

### DIFF
--- a/R/cf.R
+++ b/R/cf.R
@@ -1100,6 +1100,30 @@ num.cf <- function(cf, n=1.) {
   return (cf)
 }
 
+#' FHT ratio
+#'
+#' @description
+#' Computes Feynman-Hellmann theorem (FHT) ratio for given
+#' 2-point and 3-point cf objects.
+#'
+#' @param cf2pt `cf_orig` object.
+#' @param cf3pt `cf_orig` object.
+#' @param tau Numeric.
+#'
+#' @return
+#' returns the FHT ratio in form of a cf object.
+#' 
+#' @export
+fht_ratio.cf <- function(cf2pt, cf3pt, tau) {
+  z <- ( shift.cf(cf2pt, tau) + shift.cf(cf2pt, -tau) ) / ( mul.cf(cf2pt, 2) )
+  pre <- z / pow.cf(z*z - num.cf(z), n=0.5)
+  
+  r <- ( (shift.cf(cf3pt, tau) + shift.cf(cf3pt, -tau)) / (shift.cf(cf2pt, tau) + shift.cf(cf2pt, -tau)) - cf3pt/cf2pt )
+  res <- mul.cf( pre, (1/tau) ) * r
+  
+  return(res)
+}
+
 
 
 #' extract one single correlator object as \code{cf} object from a large
--- a/R/cf.R
+++ b/R/cf.R
@@ -1024,6 +1024,44 @@ mul.cf <- function(cf, a=1.) {
   return (cf)
 }
 
+#' Exponentiate cf objects
+#'
+#' Note that no complex arithmetic is used, real and imaginary parts are 
+#' treated as seperate and indepenent, such that the real part of cf is
+#' raised to the power of n and similarly for the imaginary part.
+#'
+#' Note that this is generally only allowed on bootstrap samples and mean values,
+#' although it makes sense in some exeptional circumstances. Don't use this
+#' function unless you're certain that you should!
+#' 
+#' @param cf `cf_orig` object.
+#' @param n Numeric.
+#'
+#' @return
+#' The value is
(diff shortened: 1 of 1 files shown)


R/cf.R
```diff
@@ -1024,6 +1024,106 @@ mul.cf <- function(cf, a=1.) {
...
+#' The value is
+#' \deqn{cf^n \,.}
+#' @export
+pow.cf <- function(cf, n=1.) {
+  stopifnot(inherits(cf, 'cf_meta'))
+  if(inherits(cf, 'cf_orig')) {
+    cf$cf <- (cf$cf)^n
+    
+    if( has_icf(cf) ){
+      stopifnot(has_icf(cf))
```
kostrzewa, line 1050: i don't think you need this check here, you've just checked it

R/cf.R
```diff
@@ -1024,6 +1024,106 @@ mul.cf <- function(cf, a=1.) {
...
+  return (cf)
+}
+
+#' Assign numeric values n to entries of cf objects
+#'
+#' Note that no complex arithmetic is used, real and imaginary parts are 
+#' treated as seperate and indepenent, such that the real part of cf is
+#' divided by itself and multiplied with n and similarly for the imaginary part.
+#'
+#' Note that this is generally only allowed on bootstrap samples and mean values,
```
kostrzewa, line 1071: i don't understand the documentation here... what is this supposed to do?

R/cf.R
```diff
@@ -1024,6 +1024,106 @@ mul.cf <- function(cf, a=1.) {
...
+#' @param n Numeric.
+#'
+#' @return
+#' The value is
+#' \deqn{cf^n \,.}
+#' @export
+pow.cf <- function(cf, n=1.) {
+  stopifnot(inherits(cf, 'cf_meta'))
+  if(inherits(cf, 'cf_orig')) {
+    cf$cf <- (cf$cf)^n
```
kostrzewa, line 1047: do you actually want to take a power of the original data or is this an operation which should only be carried out on the mean value and bootstrap samples?

R/cf.R
```diff
@@ -1024,6 +1024,106 @@ mul.cf <- function(cf, a=1.) {
...
+
+#' FHT ratio
+#'
+#' @description
+#' Computes Feynman-Hellmann theorem (FHT) ratio for given
+#' 2-point and 3-point cf objects.
+#'
+#' @param cf2pt `cf_orig` object.
+#' @param cf3pt `cf_orig` object.
+#' @param tau Numeric.
```
kostrzewa, line 1111: explain what this is please

R/cf.R
```diff
@@ -1024,6 +1024,106 @@ mul.cf <- function(cf, a=1.) {
...
+    cf$cf0 <- cf$cf.tsboot$t0
+  }
+  else cf <- invalidate.samples.cf(cf)
+  return (cf)
+}
+
+#' FHT ratio
+#'
+#' @description
+#' Computes Feynman-Hellmann theorem (FHT) ratio for given
```
kostrzewa, line 1106: a formula might help

urbach:
@NikSchlage this requires additional work!

R/cf.R
NikSchlage: Yes, because I already checked with "if(inherits(cf, 'cf_orig'))" if the cf object inherits from the class "cf_orig" which contains icf.

R/cf.R
NikSchlage: In the FHT ratio function fht_ratio.cf() I calculate the prefactor z/sqrt(z^2 - 1), i.e. pre <- z / pow.cf(z*z - num.cf(z), n=0.5). In this, z is a cf object and consequently z*z is also a cf object, because the objects are multiplied element by element. This means that you cannot simply subtract a numeric value. Therefore, I wrote the function num.cf(), which assigns a numeric value to each entry of the passed cf object, in this case 1, since "num.cf <- function(cf, n=1.)". The result is a cf object with the same dimension as z*z that can be subtracted. So here the value 1 is always subtracted. I achieve the assignment of a numeric value n to all entries via cf$cf <- n*(cf$cf/cf$cf) or cf$icf <- n*(cf$icf/cf$icf). This is what I meant with "the real part of cf is divided by itself and multiplied with n and similarly for the imaginary part.".

NikSchlage: Now I have rewritten my documentation again and hope that it is now more understandable.

NikSchlage: Should I remove the check "if( has_icf(cf) )" also in the functions num.cf()? Just like in pow.cf() don't see that it is needed.

R/cf.R
NikSchlage: Is done.

PR #310, opened 2022-07-12
Faster plotwitherror

j-ostmeyer:
`plotwitherror` takes quite a long time to plot data sets with O(100) points. In an interactive session I can actually see the errorbars emerging from left to right within several seconds. The new implementation should preserve all the functionality we had so far, but it also vectorises the call to `arrow` (responsible for drawing the errorbars) if possible, i.e. if all the points have the same color. This reduces the runtime for plotting to virtually instant.

Also, some minor uncritical changes that accumulated over time.


kostrzewa:
@j-ostmeyer Thanks. I won't really have the time to check this out soon, sorry... I agree that `plotwitherror` can be slow, but I've only ever found this to be an issue in remote X sessions. And if it's a problem there, one should use tunneled VNC or some other more efficient way of connecting to the remote X client. In any case, any speedup which doesn't break the existing functionality is very welcome, of course!

R/effectivemass.R
```diff
@@ -133,6 +134,10 @@ effectivemass.cf <- function(cf, Thalf, type="solve", nrObs=1, replace.inf=TRUE,
     if(type == "log") {
       effMass[t2] <- log(Ratio[t2])
     }
+    else if(type == "power") {
```
kostrzewa, line 137: this is good to have for the epsilon regime, thanks!

PR #314, opened 2022-11-23
support for multidimensional nlsfit [draft]

kostrzewa:
This is just a proposal for now and a real implementation would take some time and effort in order to provide the full functionality including, perhaps, even plotting support.

But relaxing the condition on the lengths of `x` and `y` (requiring the former to be divisible by the latter), `bootstrap.nlsfit` in principle immediately supports fits in more than one dimension by simply concatenating the different X variables that one wants to use and writing an appropriate fit function which indexes into these (the necessary starting indices might be passed as a vector themselves, for example).


kostrzewa:
@simone-romiti this is required to make the first fit pass

